PR #35, opened 2015-04-02
Added new metric mean absolute error

sebhahn:


sebhahn:
correction and tests are implemented, it should be ready from my side

cpaulik:
There seems to be a problem in python3. I'll check on my laptop when I have time

tests/test_metrics.py
```diff
@@ -73,3 +66,136 @@ def test_pearson_conf():
...
+    rmsd_obs = met.rmsd(x, y)
+
+    nptest.assert_equal(rmsd_obs, rmsd_pred)
+
+    # example 2, with outlier
+    x = np.arange(10)
+    y = np.arange(10) + 2
+    y[-1] = 100
+
+    rmsd_pred = np.sqrt(831)
```
cpaulik: This is not correct since the RSS is 8317 which divided by 10 should not be 831. But standard python2 division is of course dangerous and makes this work

sebhahn:
does this solve the issue for python3?

pytesmo/metrics.py
```diff
@@ -59,16 +103,19 @@ def rmsd(o, p):
...
+        Delta degree of freedom.The divisor used in calculations is N - ddof,
+        where N represents the number of elements. By default ddof is zero.
 
     Returns
     -------
     rmsd : float
         Root-mean-square deviation.
     """
-    return np.sqrt(RSS(o, p) / len(o))
+    return np.sqrt(RSS(o, p) / (len(o) - ddof))
```
cpaulik: This is actually an old bug because len(o) is a integer so division will result in an integer. Could you please add
```python
from __future__ import division
```
This should make it behave consistently in python2 and python3

sebhahn:
should work now

cpaulik:
Thank you and sorry for the old `len` bug.

sebhahn:
No problem, I'm happy that this is finished :)

Issue #46, opened 2015-06-08
AscatH25_SSM: rename read_ssm to read_ts; read should return pandas.DataFrame or pandas.TimeSeries


aplocon:

cpaulik:
Both read_ssm and read_ts are available for the moment. See https://github.com/TUW-GEO/pytesmo/commit/c8288706e27601a83bd4f67a031ae5152fe9e170

PR #47, opened 2015-06-11
Validation Framework

aplocon:


aplocon:
Also fixes #46

cpaulik:
Thank you I've merged this manually and added the examples to the `examples.rst`.
I also added back the `read_ssm` function to not break other Applications. `read_ts` is also available but just calls `read_ssm` and returns only the DataFrame

PR #75, opened 2015-12-21
Fix string concatenation for data manager

sebhahn:
Updating the string concatenation helps to avoid problems if args are e.g. one element lists or tuples. The old formatting would through an exception in these cases.


cpaulik:
Thanks.

PR #76, opened 2016-02-05
Remove grid functionality completely. Direct to pygeogrids.

cpaulik:


cpaulik:
Travis build failed because of some problem with their environment.

PR #77, opened 2016-02-05
Fix formatting in cython code. Correct docstring of boxfilter.

cpaulik:


cpaulik:
Travis build failed because grid tests still fail in this branch

PR #78, opened 2016-02-05
Fix moving average with input size 1 and add tests.

### LINKED ISSUE #74: boxcar filter fails when array of size 1 is the input

cpaulik:
Fix #74


cpaulik:
Travis failed because of https://github.com/conda/conda/issues/2028

Issue #72, opened 2015-12-04
Use pygeobase base classes for all datasets


cpaulik:
- [x] ASCAT
- [ ] ISMN
- [x] H-SAF

cpaulik:
ASCAT and H-SAF moved to https://github.com/TUW-GEO/ascat/issues/5

tracyscanlon:
ISMN is now part of another package. https://github.com/TUW-GEO/ismn

PR #81, opened 2016-03-16
Refactor H-SAF readers to use pygeobase

cpaulik:
Related to Issue #72


cpaulik:
Error in the conda dependencies makes python 3.3 build fail. Code is ok.

PR #83, opened 2016-03-22
Return more data from H07 reader.

### LINKED ISSUE #30: H07 should return more variables
Best to specify a list of variables of interest during init

cpaulik:
fix #30


cpaulik:
Build still fails because of conda issue https://github.com/ContinuumIO/anaconda-issues/issues/686

PR #118, opened 2017-06-20
add respect leap years for climatology

iteubner:


iteubner:
I like working with the anomaly and climatology calculation in pytesmo - makes life easier. For climatologies of MODIS data, however, I get double entries due to leap years. I added the leap year option, which is already present in the anomaly function, also to the climatology function and set the default to True to be consistent with the anomaly function. However, the current version would rather correspond to False as the default.

cpaulik:
I had made the comments that just appeared when you submitted the PR but forgot to press the button to publish them sorry for that.

cpaulik:
Thanks.

cpaulik:
This is now part of version v0.6.7

PR #131, opened 2018-03-26
Readme

mazweg:


sebhahn:
@cpaulik are you happy with the new readme format for pytesmo?

cpaulik:
I've pushed one more commit adding the meaning of the pytesmo name into the first sentence. Otherwise I'm happy with it. Thanks.

PR #132, opened 2018-03-28
use ascat0.9 in travis

mazweg:
@cpaulik  this will fix the tests. 
In order to use ascat v0.10 further changes are needed.


cpaulik:
I'd rather keep the tests failing until we have a proper solution. So we don't forget that something is wrong.

cpaulik:
Superseded by #133

PR #133, opened 2018-03-28
Update test data repository and tests to be compatible with latest ascat

cpaulik:
…cat release


cpaulik:
If this problem occurs again we should probably look into how we can use the ascat-test data repository directly.

@sebhahn please limit format changes to a minimum :smile:

sebhahn:
yes I know, sorry :smiling_imp:

PR #137, opened 2018-06-04
Linear regression scaling with nans

wpreimes:
In this implementation, when there are nans in the src or ref series, these lines are removed from the data frame (pandas dropna) (as stats.linreg needs arrays of equal lengths), when the correction params are found (slope, inter) they are applied back to the full array (until now stats.linreg raised an error). Bias correction will probably be worse if there are a lot if nans, but the function should not fail anymore.
@cpaulik : You think this is ok?

### DIFF
--- a/pytesmo/scaling.py
+++ b/pytesmo/scaling.py
@@ -172,8 +172,10 @@ def linreg(src, ref):
         dataset scaled using linear regression
     """
 
+    df = pd.DataFrame(data={'src': src, 'ref': ref}).dropna()
+
     slope, intercept, r_value, p_value, std_err = stats.linregress(
-        src, ref)
+        df.src, df.ref)
 
     return np.abs(slope) * src + intercept
 
--- a/pytesmo/scaling.py
+++ b/pytesmo/scaling.py
@@ -155,6 +155,54 @@ def min_max(src, ref):
             (np.max(ref) - np.min(ref)) + np.min(ref))
 
 
+def linreg_stored_params(src, slope, intercept):
+    '''
+    Scale the input data with passed correction values
+
+    Parameters
+    ----------
+    src : numpy.array
+        Candidate values, that are scaled
+    slope : float
+        Multiplicative correction value
+    intercept : float
+        Additive correction value
+
+    Returns
+    -------
+    src_scaled : numpy.array
+        The scaled input values
+    '''
+
+    return np.abs(slope) * src + intercept
+
+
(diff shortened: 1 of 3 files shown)


pytesmo/scaling.py
```diff
@@ -172,8 +172,10 @@ def linreg(src, ref):
         dataset scaled using linear regression
     """
 
+    df = pd.DataFrame(data={'src': src, 'ref': ref}).dropna()
```
cpaulik: I see the `linreg` and similar functions as pure functions that do not care about the input a lot. Any user can do removing of nan values outside of this function. A general solution like this is problematic since there can always be custom no data values like e.g. `-99999` that we can not remove automatically.

wpreimes: ok. makes sense, then we could - as you already suggested - try to split it up in 2 functions:
- ```linreg_params(src, ref)``` -> returns _slope_ and _inter_ correction values (for a common subset of the data that I want to scale, must be done beforehand, will fail if there are NaNs )
- ```linreg_adjust(src, slope, inter)``` -> applies slope, inter on (all, passed) values, returns bias-corrected values.

Then one could use them separately if _src_ and _ref_ don't match to not lose values before calling the scaling function (and having not all values scaled). And ```linreg(src, ref)``` would combine them (as already by now, but would still fail with NaNs in _src_  and _ref_  (or of course if they have different lengths)). But seems quite complicated, I would just use scipy then from the start :-)
But we could also identify NaNs in the higher-level ```scale()``` function (eg. add another parameter), that would handle all that when calling ```linreg()``` or ```mean_std()``` then in ```scale()```. Can't say much about the CDF matching though...
Don't know if this makes sense to you, I just found the mean_std and linreg scaling from pytesmo not that useful for my applications until now for the described reasons (I basically never have continuous/matching TS).

cpaulik: Yes we can split it into two functions. 

Handling NaN values in a general way is too difficult, I would let the user do that.

For getting matching time series I would just use the [temp_matching](https://github.com/TUW-GEO/pytesmo/blob/master/pytesmo/temporal_matching.py#L11) module which will drop NaN as defined by pandas.

wpreimes:
i split the linreg function to fix #137 as discussed in the issue

cpaulik:
That looks good. Could you update the `CHANGES.rst`? Then we can merge it. Just add a new line with `unreleased` on top of the changelog file.

PR #124, opened 2017-10-20
Remove ISMN module and change documentation to point to ismn package.

cpaulik:

PR #139, opened 2018-07-02
Fix examples

awst-baum:
Fixing outdates imports and testdata references in examples.


cpaulik:
Thanks.

I think this is a good start for making you familiar with the code and for updating the examples. 

Since we now have an ascat package. We should remove 

- `Read_H_SAF_images.py`
- `plot_ASCAT_data.py`
- `read_ASCAT_H25.py`

And even the `plot_ISMN_data.py` should probably removed in https://github.com/TUW-GEO/pytesmo/pull/124

@awst-baum Also asked via email if there is an automated way of getting the examples into the documentation. In the past we have used Ipython notebooks for that. See for example the [notebook about triple collocation](https://github.com/TUW-GEO/pytesmo/blob/master/docs/Triple%20collocation.ipynb) . From these notebooks we can export both `.rst` and `.py` files. But nowadays we can probably just include the notebooks in the examples folder since they are very common and github also renders them nicely.

cpaulik:
I will merge this and we can then start the conversion to notebooks that are rendered by sphinx in a separate branch.

PR #140, opened 2018-07-18
Add masking helpers to validation framework

awst-baum:
Extends the existing MaskingAdapter so that it can be used reduce multi-column masking datasets to one column and do "not equal" comparisons. Adds a SelfMaskingAdapter that wraps a data reader and allows to mask the data while reading, based on a comparison operation on one column (column, operation, and threshold can be specified).

### DIFF
--- a/pytesmo/validation_framework/adapters.py
+++ b/pytesmo/validation_framework/adapters.py
@@ -54,23 +54,30 @@ class MaskingAdapter(object):
 
     """
 
-    def __init__(self, cls, op, threshold):
+    def __init__(self, cls, op, threshold, column_name = None):
         self.cls = cls
 
         self.op_lookup = {'<': operator.lt,
                           '<=': operator.le,
                           '==': operator.eq,
+                          '!=': operator.ne,
                           '>=': operator.ge,
                           '>': operator.gt}
         self.operator = self.op_lookup[op]
         self.threshold = threshold
 
+        self.column_name = column_name
+
     def read_ts(self, *args, **kwargs):
         data = self.cls.read_ts(*args, **kwargs)
+        if self.column_name is not None:
+            data = data.loc[:, [self.column_name]]
         return self.operator(data, self.threshold)
 
     def read(self, *args, **kwargs):
         data = self.cls.read(*args, **kwargs)
+        if self.column_name is not None:
+            data = data.loc[:, [self.column_name]]
         return self.operator(data, self.threshold)
 
 
--- a/pytesmo/validation_framework/adapters.py
+++ b/pytesmo/validation_framework/adapters.py
@@ -117,8 +117,7 @@ class SelfMaskingAdapter(object):
 
     def __mask(self, data):
         mask = self.operator(data[self.column_name], self.threshold)
-        data = data[mask]
-        return data
+        return data[mask]
 
     def read_ts(self, *args, **kwargs):
         data = self.cls.read_ts(*args, **kwargs)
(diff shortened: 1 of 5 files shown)


pytesmo/validation_framework/adapters.py
```diff
@@ -39,40 +39,93 @@
...
+    Parameters
+    ----------
+    cls: object
+        has to have a read_ts or read method
+    operator: string
+        one of '<', '<=', '==', '>=', '>', '!='
+    threshold:
+        value to use as the threshold combined with the operator
+    column_name: string
+        name of the column to cut the read masking dataset to
```
cpaulik: I guess this should be something like `name of the column to apply the threshold to`?

pytesmo/validation_framework/adapters.py
```diff
@@ -39,40 +39,93 @@
...
         data = self.cls.read_ts(*args, **kwargs)
-        return self.operator(data, self.threshold)
+        return self.__mask(data)
 
     def read(self, *args, **kwargs):
         data = self.cls.read(*args, **kwargs)
-        return self.operator(data, self.threshold)
+        return self.__mask(data)
+
+class SelfMaskingAdapter(object):
```
cpaulik: Could you add a small paragraph to the documentation about the `SelfMaskingAdapter` and the differences to the `MaskingAdapter`. I think that would help us in making it clearer what they both do. I think we might be able to combine them into one class since they are very similar. What do you think?

pytesmo/validation_framework/adapters.py
awst-baum: Yes, that sounds better (although the given column will be the only one remaining in the DataFrame after masking).

pytesmo/validation_framework/adapters.py
awst-baum: Re documentation, sure, as soon as we've sorted out what we actually want to do. Re combining: pro: avoiding code repetition, contra: both adapters do something different and they yield a quite different result. My personal taste as a dev would be to have two separate classes for two separate concerns - but I could derive them from a common base class to avoid code repetition. If I do that, should I use an https://docs.python.org/2/library/abc.html ?

pytesmo/validation_framework/adapters.py
awst-baum: Are we talking about the changes in MaskingAdapter or about the SelfMaskingAdapter?

pytesmo/validation_framework/adapters.py
cpaulik: For the documentation I think we should edit the ipython notebooks and automatically include them in the sphinx documentation. For now editing https://github.com/TUW-GEO/pytesmo/blob/master/docs/validation_framework.ipynb and exporting it as rst should also work.

cpaulik: I think we can keep the classes separate for now.

pytesmo/validation_framework/adapters.py
cpaulik: SelfMaskingAdapter. Line 101 of the diff.

awst-baum: Oh, sorry, my bad. I've changed it.

pytesmo/validation_framework/adapters.py
awst-baum: OK, I've updated the docs - what do you think?

cpaulik: Looks good, I'll merge this after a quick rebase if all the tests still pass.

Issue #136, opened 2018-05-25
Fix issues related to pandas 0.23


sebhahn:
Some tests are failing because pandas 0.23 is more restrictive, I think it shouldn't be a lot of effort to fix them.

https://pandas.pydata.org/pandas-docs/stable/whatsnew.html

cpaulik:
Yes we should fix them. Did you check which change exactly broke pytesmo?

sebhahn:
Not yet, but @mazweg fixed one of the errors for the ismn package https://github.com/TUW-GEO/ismn/commit/67e318a3ae845548fe3add588a64db0eb65dffe5

awst-baum:
I've tried my hand at this but it's not quite clear to me what change in pandas brought the problems about. See my description in the pull request.

PR #141, opened 2018-07-19
Try fixing failing issues with new pandas version (0.23)

awst-baum:
This is my attempt at fixing #136 

These two one-line-changes make the unit tests pass. WARNING: However, I'm not sure if I completely understand what the code is doing and what changed.

My understanding is that the problems come from two errors at two places in the code:

pytesmo/validation_framework/temporal_matchers.py:70
E   KeyError: "[('index', '')] not found in axis"

and

pytesmo/validation_framework/validation.py:270
E   AttributeError: 'Index' object has no attribute 'droplevel'

I guess they occur because Pandas changed the way it creates column names. In the matcher case, the column name is ('index', '') with pandas 0.22 and just 'index' with pandas 0.23. In the validation case, pandas 0.22 produces a MultiIndex while pandas 0.23 produces a simple Index with tuples as column names.

The changes I propose make pytesmo work with pandas 0.23 - but NOT with 0.22 (not downward compatible). To do that, one would have to investigate the column name creation, I guess.

I hope this helps...

### DIFF
--- a/pytesmo/validation_framework/temporal_matchers.py
+++ b/pytesmo/validation_framework/temporal_matchers.py
@@ -35,7 +35,7 @@ import itertools
 import pytesmo.temporal_matching as temp_match
 
 import pandas as pd
-
+from distutils.version import LooseVersion
 
 class BasicTemporalMatching(object):
     """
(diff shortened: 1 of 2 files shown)


cpaulik:
Thanks. I think it is ok to update our requirement to pandas 0.23. @sebhahn Do you foresee any problems?

Another option would be to check `pandas.__version__` but I'm not sure that is worth it.

sebhahn:
I'd suggest to check `pandas.__version__` since it is not much to change

awst-baum:
Here's my attempt at making the change downward compatible.

sebhahn:
I'm not sure if users are ready to change to the latest pandas version and since pytesmo is more famous among other packages from our group I'd suggest supporting multiple pandas versions at least for some time.

awst-baum:
Whatever you want guys, just make a decision ;-)

cpaulik:
ok. I'll merge this and also fix the python 3.6 test

Issue #138, opened 2018-06-29
bias calculation


wpreimes:
Hey! Bias is currently np.mean(observations) - np.mean(predictions), but should be the other way around, right? @iteubner also mentioned this some time ago.

sebhahn:
Well, it's always a matter of how you pass the arguments to the bias function. In the paper of [Janssen et al.](https://www.sciencedirect.com/science/article/pii/0304380095000849) it is defined as `predictions - observations`.

Anyway, I think it is more plausible with `observations - reference`

wpreimes:
So you would keep `predictions - observations`? Why not stick with the definition from the paper?

sebhahn:
I think the documentation should be updated and make it clear, because in many cases we all work with "observations" only and not with model data.

In case someone has many data sets (x1, x2, x3, etc.) and a reference data set (y) (no matter if xi or y are model data, observations, etc.) it should be clear that it is always xi - y. And than the arguments should have a meaningful documentation.

wpreimes:
So would make sense do give them more general names like `can` (candidate) or `src`, and `ref` (reference) and define bias as `mean(can)`- `mean(ref)`? Would you do this also for the other metrics?

wpreimes:
sure, but I'd suggest having consistent names, even if it doesn't make a difference computationally. If you want, I can updated the names and make a PR with the suggested changes.

cpaulik:
I would not like to switch the argument order since this would lead to all existing code suddenly producing different results.

As far as I've seen by quickly going through the code bias is the only metric we have where the argument order matters.

About the documentation. I've actually liked the way it was before `o` and `p` where introduced where we just had `x` and `y` and the calling function has to decide in which direction the bias is interesting based on the application. E.g. the validation framework [call the bias function with `bias(ref, other)`](https://github.com/TUW-GEO/pytesmo/blob/master/pytesmo/validation_framework/metric_calculators.py#L116).

iteubner:
You are right, it is only a matter of how the function is called. But I still think it would be more intuitive for someone who takes a look at the code to see the reference (usually called observations) as the second argument in the subtraction, i.e. mean(p)-mean(o). 

Maybe I am missing something, but if one would exchange both the order of the arguments and the order of the subtraction it should stay the same for existing code, right?

Alternatively, one could maybe state that the predictions are the reference.

cpaulik:
[...]

PR #146, opened 2018-07-30
Change argument names for metrics functions

wpreimes:
Set the names for metrics input data to general names (`x` and `y` instead of `o` and `p`), so the bias function can not be misinterpreted anymore.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 unreleased 
 ===================
 
+- Rename arguments for metrics functions to be more general
 - Split linreg scaling function to determine and apply corrections separately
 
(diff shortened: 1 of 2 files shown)


wpreimes:
this should fix #138

sebhahn:
Looks good. However, I'd rather prefer another description of `x, y`. 

E.g. instead of `First set of values` use `First input vector` and `Second input vector`

wpreimes:
Done

sebhahn:
:+1:

cpaulik:
Now the bias is different from the rest. Was that on purpose?

wpreimes:
oops, no, I will change it

PR #156, opened 2019-03-12
add intercomparison metrics calculator and tc for qa4sm

wpreimes:


wpreimes:
@tracyscanlon maybe before merging this, lets discuss it with the others, also I might add another commit here.

tracyscanlon:
@wpreimes Let's discuss with the others. And yes, it will need testing.

pytesmo/validation_framework/metric_calculators.py
```diff
@@ -236,3 +240,725 @@ def calc_metrics(self, data, gpi_info):
...
+
+            if self.calc_tau:
+                dataset['{:}_tau'.format(tds_name_key)][0] = tau
+                dataset['{:}_p_tau'.format(tds_name_key)][0] = p_tau
+
+        return dataset
+
+
+
+if __name__  == '__main__':
```
sebhahn: this can be removed

wpreimes: ok, thanks, @sebhahn have you noticed that here the wrong R is used? https://github.com/wpreimes/pytesmo/blob/d715209417c6165f599d7e9848b7ca10b52a2b19/pytesmo/validation_framework/metric_calculators.py#L482

sebhahn: No, could you fix this? Is not a big problem though, because we do not use spearman

wpreimes: sure, already fixed it

wpreimes:
I have also removed the old ISMN code and added the ISMN reader package as a dependecy (same as in #124 , which can be closed after merging this as well) and updated the tests accordingly

sebhahn:
Is this pr ready?

wpreimes:
Yes you can merge it from my side

tracyscanlon:
No objections here. Please go ahead and merge.

cpaulik:
Two small comments:

I don't think we should add `ascat` and `ismn` as requirements of pytesmo. This somewhat defeats the purpose of splitting them into separate packages if they are not installable separately. We could put them into the `extras` section of `setup.cfg` then somebody could e.g. install `pip install pytesmo[ascat,ismn]` and get all dependencies that are required to work with those datasources. We could also make an extra `dataset-readers` which could include `ascat,gldas,ismn,ecmwf...` so installing `pip install pytesmo[dataset-readers]` would install all of them.

The second comment/question is about the `build_filename` method that limits the filename to 144 characters. Where did the 144 come from? On Linux the filename has a limit of 255 bytes so there it should not be a problem. On Windows the whole path can not be longer than 255 characters so there we might even have a problem with 144 if the basepath is too long.

wpreimes:
Good ideas. We will look into this! Yes the filename function needs some rework, the 144 characters came from my windows machine (as win counts the whole path as you said). I will think of a better solution.

PR #158, opened 2019-04-29
Add Python 3.7 and remove 3.5 from CI

cpaulik:
Update CI a bit.

Dropped Python 3.5 testing and added 3.7. 

For Windows build I had to drop older 32 bit versions since the conda packages of the necessary netcdf4 version do not exist. But I'm not sure if anybody is actually using a 32 bit system.


sebhahn:
`pip install pybufr-ecmwf` - is this working well via pip?

cpaulik:
Seems to work much better than in the past. We can switch back to conda as soon as https://github.com/conda-forge/pybufr-ecmwf-feedstock/pull/7 is fixed. 

But for now it is fine. Especially since pybufr-ecmwf is not really needed by pytesmo itself. Only for one specific resampling test which we could replace if somebody has time.

PR #159, opened 2019-04-29
Added fillna and min_obs options to moving average

teije01:
as said in title + added test functions for it


cpaulik:
We should also change the `climatology` function to use these new keywords. 

I think the default in the climatology calculation can be `fillna=True` since it will not change the output of anomaly calculations. The only difference is that climatologies will be filled if they have no data values.

@sebhahn @tracyscanlon Any worries if we change the default behavior of climatologies to fill no data values?

teije01:
This was added in the latest commit. This does change the default behaviour, so be adviced

wpreimes:
Can you give us some more details on what you want to do?

cpaulik:
> Can you give us some more details on what you want to do?

The current problem is that if you pass a time series with `no data` values into `moving_average` it will not fill any `no data` gaps with values. Even thought that is what most people would expect. So this should definitely be an option if not the default behavior.

This behavior leads to strange results when calculating a climatology.

e.g. calling `calc_climatology` with `moving_avg_orig=5` on a time series with `no data` gaps will give a climatology with `no data` gaps even though the original input series was smoothed.

teije01:
To add to this,

The behaviour of the moving average remains unchanged, though new functionality will be added with these changes. 

For the climatology, this new functionality will be added as well which enables us to fill no-data gaps which you generally don't expect when calculating a climatology or moving average (just like with the native pandas `Series.rolling().mean()` calculation. However, the current default behaviour in the climatology (not filling nan-values) is different from the ones in these proposed changes (filling nan-values).

Still, nothing would change for datasets that are continuous, though for nan-filled datasets, there will be a difference.

For the reason of not changing the current behaviour, one could argue to set the default to not fill nan-values in the `calc_climatology`, that is also in line with the `moving_average`. However, one could also argue to change both defaults to `True` to be both consistent within the package, and to get the expected behaviour of a moving_average function. However, this affects anyone who is currently using these scripts.

Hopefully we can agree on one of three default settings:
- not changing any of the previous settings
- consistency 
- mixed

NexusMaxus:
To get rid of the spike on the leapday created if climate smoothing is not applied, which if a sample taken of 4 years for example would be based on one value only, I added an interpolation for the leapday based on the day before and after in the climatology. It is now optional, but I would say it should be the default.

cpaulik:
We should add some tests for the new climatology keywords.

I will clean up the commits and the formatting in the end when the functionality is ready.

cpaulik:
Tests in travis fail because of https://github.com/Unidata/netcdf4-python/issues/919

cpaulik:
We would like to merge this after some cleanup. So if nobody has any issues with this I would go ahead later this week.

@sebhahn @tracyscanlon last chance

sebhahn:
Fine for me!

sebhahn:
@cpaulik: Before merging this PR and creating a new release (e.g. v0.7.0), could you please also create a new release before (e.g. v0.6.11)? Thanks!

cpaulik:
> @cpaulik: Before merging this PR and creating a new release (e.g. v0.7.0), could you please also create a new release before (e.g. v0.6.11)? Thanks!

Yes, no problem

cpaulik:
I've released `v0.6.11` without these changes and `v0.7.0` from the latest master. Windows builds are still ongoing and will be added to pypi as they finish

PR #160, opened 2019-06-12
QA4SM updates

wpreimes:
- Update Intercomparison variable names
- Add RSS to intercomparison
- Start with new results manager


cpaulik:
About dependencies. If you agree then I think it would be best if we make some dependencies optional.

e.g. `pip install pytesmo[validation_framework]` would install everything.
`pip install pytesmo` alone would only install dependencies that are necessary to use pytesmo as a library.

I can work on that or show you how it is done if you don't know.

The main reason is that I think we should avoid heavy things like `xarray` if they will not be used in a lot of cases.

sebhahn:
Looking forward to see how this works :rocket:

wpreimes:
@cpaulik I will look at this and see how far I get ;)

wpreimes:
I have taken out the infinished stuff from this PR (results manager) and collected multiple changes from different branches instead (e.g. from @tracyscanlon). We can merge this now and I will open separate PRs for the remaining changes.

.gitignore
```diff
@@ -48,3 +48,5 @@ MANIFEST
 
 # ignore ISMN metadata files as they are different on win/linux
 */python_metadata/*
+
+local_scripts/*
```
tracyscanlon: Is this necessary for the package?

requirements.txt
```diff
@@ -8,4 +8,6 @@ pygeobase>=0.3.15
 pynetcf>=0.1.17
 ascat>=1.0
 ismn>=0.3
-configparser
\ No newline at end of file
+configparser
+pykdtree
+xarray
```
tracyscanlon: is xarray used without the new results manager?

pytesmo/temporal_matching.py
```diff
@@ -3,7 +3,7 @@
 """
 
 import numpy as np
-from scipy.spatial import cKDTree
+from pykdtree.kdtree import KDTree
```
tracyscanlon: please state in the release notes why we have changed from scipy to pykdtree

tests/test_validation_framwork/test_metric_calculators.py
```diff
@@ -4,30 +4,55 @@
 from datetime import datetime
 from pytesmo.validation_framework.metric_calculators import *
 
+import warnings
```
tracyscanlon: What warnings are we filtering for here? Should they be ones that are ignored?

.gitignore
wpreimes: No it's just for me :) Any for anyone else who wants a location in the package to collect test files that should not be tracked

requirements.txt
wpreimes: no, removed it

pytesmo/temporal_matching.py
wpreimes: it's faster

wpreimes:
@tracyscanlon Point 2), where does it fail, it passes in Travis https://travis-ci.org/TUW-GEO/pytesmo/jobs/579909814#L1198

tracyscanlon:
> 
> 
> @tracyscanlon Point 2), where does it fail, it passes in Travis https://travis-ci.org/TUW-GEO/pytesmo/jobs/579909814#L1198

It fails if you dont have pyburf-ecmwf in the environment. And in the ASCAT package there is a try except which allows for this not being installed. So we can either insist that pybufr-ecmwf is part of the pytesmo environment or we should write a work-around for this test if pybufr-ecmwf is not in the environment.

wpreimes:
Well it's optional (`optional_requirements.txt`), I think it should be in the `environment.yml`

sebhahn:
pytesmo should not take care of readers. If ASCAT data in bufr format are read, than this should come from the ascat package

tracyscanlon:
@sebhahn would you then take care of this requirement in ascat? I assume the reason it is not already a requirement of ascat is due to the size of the pybufr-ecmwf package.

sebhahn:
I just checked the package and I can recall that the problem was (and probably still is) that pybufr-ecmwf can only be installed correctly using conda, because of its C library dependencies

tracyscanlon:
In which case, we should alter the test to work around this and neither package will be able to ensure that pybufr-ecmwf is installed

wpreimes:
> 
> 
> In which case, we should alter the test to work around this and neither package will be able to ensure that pybufr-ecmwf is installed

I will do that

wpreimes:
Done, pybufr-ecmwf is not used anymore and the tests now uses generic test data instead of the ascat reader. And I made a little detour into cdf_matching, which we should keep in mind for the next release (#175)

tracyscanlon:
Cloned, made environment and tested. All works so merging.

PR #161, opened 2019-06-22
copy mask in resample_to_grid(), thus preventing it to get modified during iterations

Lmoesinger:
**Issue:** Currently `mask` is used to generate a masked array each loop. However, if `data` in 270: `output_array[~mask] = data` contains nan values, the mask of `output array` will change, and by extensions also the original `mask`. Therefore in subsequent  loop iterations `mask` will be different.

**Fix:** Just passing a copy of `mask` prevents it from changing between iterations.

### DIFF
--- a/pytesmo/grid/resample.py
+++ b/pytesmo/grid/resample.py
@@ -266,7 +266,7 @@ def resample_to_grid(input_data, src_lon, src_lat, target_lon, target_lat,
                 target_lat.shape, dtype=orig_data.dtype) + fill_value
         else:
             output_array = np.zeros(target_lat.shape, dtype=orig_data.dtype)
-            output_array = np.ma.array(output_array, mask=mask)
+            output_array = np.ma.array(output_array, mask=mask.copy())
         output_array[~mask] = data
 
         output_data[param] = output_array.reshape(output_shape)


cpaulik:
Would the `hard_mask` keyword also solve it? See https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.ma.array.html#numpy.ma.array

Lmoesinger:
I think it would in my case (cant check right now), but this alters the properties of `output_data`, and therefore could cause problems for projects that are already using this function.

cpaulik:
We can always call `soften_mask` on the output masked arrays to make them the same as they are now. 

In any case. Do you have a small testcase that we could include that illustrates the problem. Since this is such a hidden thing a test would be nice.

Lmoesinger:
I had the issue when data was a dict of numpy.ma.arrays. Is the function even meant to work with those, or is the goal to just support standard numpy.arrays?

cpaulik:
The function is defined for normal numpy arrays with all masked data points removed from the data itself and from the coordinate arrays.

If you need to add the data back into a bigger array after resampling then this is best handled outside of this function

Lmoesinger:
hm, i have trouble creating a test that breaks (with or w/o using masked arrays). Anyway, filling them before resampling solves the issue, so im gonna close it since it is not a bug.

A bit offtopic, but in my fork `resample_to_grid()` properly works with masked arrays as far as i can tell. In the master branch the function takes a numpy array but returns a masked array. It would be more intuitive if both input and output have the same class IMHO. Alternatively one could also fill the arrays in `output_data` before returning them, to keep it consistent. But this is a design choice that i leave to you.

cpaulik:
I think this behavior is documented sufficiently in the `fill_value` parameter. If no `fill_value` is given we have no other choice than to return a masked array that the user can then fill however she sees fit.

Thanks anyway for taking a detailed look at the function

Issue #150, opened 2018-08-28
Temporal matcher can't handle timezone information in DataFrame indices


awst-baum:
The test that reproduces the issue:

    def test_timezone_handling():
        data = np.arange(5.0)
        data[3] = np.nan

        match_df = pd.DataFrame({"data": data}, index=pd.date_range(datetime(2007, 1, 1, 0),
                                                                "2007-01-05", freq="D", tz="UTC"))
        timezone = pytz.timezone("UTC")
        ref_df = pd.DataFrame({"matched_data": np.arange(5)},
                                index=[timezone.localize(datetime(2007, 1, 1, 9)),
                                    timezone.localize(datetime(2007, 1, 2, 9)),
                                    timezone.localize(datetime(2007, 1, 3, 9)),
                                    timezone.localize(datetime(2007, 1, 4, 9)),
                                    timezone.localize(datetime(2007, 1, 5, 9))])
        matched = tmatching.matching(ref_df, match_df)

        nptest.assert_allclose(np.array([0, 1, 2, 4]), matched.matched_data)
        assert len(matched) == 4

The exception that occurs:

    tests/test_temporal_matching.py:61: in test_timezone_handling
        matched = tmatching.matching(ref_df, match_df)
    pytesmo/temporal_matching.py:157: in matching
        matched_data = matched_data.join(match)
    .../python2.7/site-packages/pandas/core/frame.py:5316: in join
        rsuffix=rsuffix, sort=sort)
    .../python2.7/site-packages/pandas/core/frame.py:5331: in _join_compat
        suffixes=(lsuffix, rsuffix), sort=sort)
    .../python2.7/site-packages/pandas/core/reshape/merge.py:58: in merge
        return op.get_result()
    .../python2.7/site-packages/pandas/core/reshape/merge.py:582: in get_result
        join_index, left_indexer, right_indexer = self._get_join_info()
    .../python2.7/site-packages/pandas/core/reshape/merge.py:736: in _get_join_info
        sort=self.sort)
    .../python2.7/site-packages/pandas/core/indexes/datetimes.py:1075: in join
        this, other = self._maybe_utc_convert(other)
    .../python2.7/site-packages/pandas/core/indexes/datetimes.py:1084: in _maybe_utc_convert
        raise TypeError('Cannot join tz-naive with tz-aware '
    E   TypeError: Cannot join tz-naive with tz-aware DatetimeIndex

The root of the problem: pytesmo.temporal_matching.df_match, line 85:

[...]

PR #164, opened 2019-07-11
Change adapters so that they can handle ascat timeseries and timezones

awst-baum:
Ascat returns data not in a dataframe but in an ascat timeseries object. This produces an exception if pytesmo tries to access the data's columns. This PR works around that.

Also, if data readers add timezone information to data, this PR normalizes to UTC and then drops tz info.


awst-baum:
@cpaulik @sebhahn Can you comment on if and how this PR would interfere with timezone handling and additional information about TS in pytesmo?
Can we pull this in until the other changes are made?

sebhahn:
Could you try the new temporal matching e.g. https://github.com/TUW-GEO/pytesmo/pull/176/commits/69d33fca3574bc923a3bfc5984e34bf24e6f3867 in the pull request #176

sebhahn:
There shouldn't be any conflict/complain about the timezone during the temporal matching, but maybe the timezone is also an issue somewhere else

awst-baum:
See issue https://github.com/TUW-GEO/pytesmo/issues/150

sebhahn:
the test you've provided in #150 works without a problem using #176

awst-baum:
Cool! Can we then add that test back into the pytesmo tests?

sebhahn:
done 3cd2fb0

awst-baum:
Great, thanks! Once your changes go to the master branch,  `__drop_tz_info` can be removed from this PR. Until it does, I'll leave it in there...

pytesmo/validation_framework/adapters.py
```diff
@@ -34,9 +35,39 @@
 import operator
 from pytesmo.time_series.anomaly import calc_anomaly
 from pytesmo.time_series.anomaly import calc_climatology
+from pandas import DataFrame
 
 
-class MaskingAdapter(object):
+class BasicAdapter(object):
+    """
+    Base class for other adapters that works around the ascat reader not
```
tracyscanlon: Not just for ascat. Would work for any data in this format.

pytesmo/validation_framework/adapters.py
```diff
@@ -34,9 +35,39 @@
...
+    returning a DataFrame. Also removes unnecessary timezone information in data.
+    """
+    def __init__(self, cls):
+        self.cls = cls
+
+    def __get_dataframe(self, data):
+        if ((not isinstance(data, DataFrame)) and (hasattr(data, 'data')) and (isinstance(data.data, DataFrame))):
+            data = data.data
```
tracyscanlon: could add a key word argument for the name of the 'data', i.e. (hasattr(data, keyword_name_of_data)). Then if not named 'data', this could be overwritten. May be useful if a reader returns more than one data frame with data in it.

pytesmo/validation_framework/adapters.py
```diff
@@ -34,9 +35,39 @@
...
+        return data
+
+    def __drop_tz_info(self, data):
+        if data.index.tz is not None:
+            warnings.warn('Dropping timezone information for data')
```
tracyscanlon: Should we put in what the timezone was?

awst-baum:
> Should we put in what the timezone was?

How do you mean?
(Hopefully, the timezone bit can be removed anyway when Sebastian's changes get merged anyway, but still)

pytesmo/validation_framework/adapters.py
tracyscanlon: i mean stating the timezone that has been dropped, i.e. UTC. May be a clearer message to the user. For example if they didn't know the timezones on the data going in, it would now tell them and may help them resolve errors more easily. Only a suggestion.

awst-baum:
OK, suggestions implemented.

wpreimes:
@tracyscanlon or @sebhahn can you merge this please, everything ok from my side (even the windows builds work now), then #177 can be closed as well.

cpaulik:
I will look at the temporal matching merge request this week. Then we can hopefully merge it and remove the time zone dropping from this one.

awst-baum:
I've merged it now because:
- this branch contains the fixes for AppVeyor as well (and we want those)
- it's not hard to deactivate/revert/change this change again
- in case temporal matching takes a little longer, we've got a workaround now

Issue #147, opened 2018-08-08
Revisit temporal collocation and add collocation examples to documentation


sebhahn:
Temporal collocation should be revisited (e.g. performance, features) and more tests should be added. In particular, more information should be provided how exactly it works and how duplicates are treated. Maybe a plot summarizing temporal collocation information/results would be already a good step forward. 

`drop_duplicates=True` - users need to understand the impact of turning duplicates on, which means from a scatterplot point of view several y-values have the same x-values, which can have a positive/negative impact on the overall metric depending on the x-position and spread of y-values (e.g. linear regression). Furthermore, duplicates might change depending on the reference, because matching a 12-hour sampled data set with a 3-hour sampled data set leads to different results depending on the selected reference data set.

Special strategy for matching soil moisture data sets? I haven't investigated the impact so far, but for soil moisture it probably makes sense to use a certain strategy during temporal matching. E.g. Matching the reference data set with past observations may cause a mismatch because a rain event could have happened during both observations and only the 2nd data sets has captured it. Unfortunately, the same could happen if the future observation is used for matching. I think it is not so easy and some tests are necessary to decide how to match in the best way, because even two measurements are close (e.g. 6 hours) in time you cannot rule out that a rain event was in between, or if both should have seen it but the soil moisture retrieval is wrong/bad.

In summary, time matters and we should make users aware of that during the computation of metrics.

s-scherrer:
Closed via #204

PR #176, opened 2019-07-15
[WIP] Fix performance issue of temporal matching, add new matching func

### LINKED ISSUE #169: Replace df_match by new/faster implementation
`pytesmo.temporal_matching.df_match` can be implemented with a faster approach using `pandas.merge_asof`.

At the same time we can also drop the `asym_window` keyword since it was only used for CCI IDL backward compatilibity

sebhahn:
- [x] Fix #169 
- [x] Test timezone information #150
- [x] Update documentation #147
- [ ] Discuss return value (tuple of dataframes or merged dataframe) of new temporal matching function

### DIFF
--- a/tests/test_temporal_matching.py
+++ b/tests/test_temporal_matching.py
@@ -1,5 +1,4 @@
-# Copyright (c) 2015,Vienna University of Technology,
-# Department of Geodesy and Geoinformation
+# Copyright (c) 2019, TU Wien, Department of Geodesy and Geoinformation
 # All rights reserved.
 
 # Redistribution and use in source and binary forms, with or without
@@ -45,19 +44,29 @@ def test_df_match_borders():
 
     See issue #51
     """
+    ref_df = pd.DataFrame(
+        {"data": np.arange(5)}, index=pd.date_range(
+            datetime(2007, 1, 1, 0),
+            "2007-01-05", freq="D"))
 
-    ref_df = pd.DataFrame({"data": np.arange(5)}, index=pd.date_range(datetime(2007, 1, 1, 0),
-                                                                      "2007-01-05", freq="D"))
     match_df = pd.DataFrame({"matched_data": np.arange(5)},
                             index=[datetime(2007, 1, 1, 9),
                                    datetime(2007, 1, 2, 9),
                                    datetime(2007, 1, 3, 9),
                                    datetime(2007, 1, 4, 9),
                                    datetime(2007, 1, 5, 9)])
+
     matched = tmatching.df_match(ref_df, match_df)
 
     nptest.assert_allclose(
         np.array([0.375, 0.375, 0.375, 0.375, 0.375]), matched.distance.values)
+    nptest.assert_allclose(np.arange(5), matched.matched_data)
+
+    matched = tmatching.df_temp_merge(ref_df, match_df, return_distance=True)
+    dist = matched['dist_other_0'].values / np.timedelta64(1, 'D')
+
+    nptest.assert_allclose(
+        np.array([0.375, 0.375, 0.375, 0.375, 0.375]), dist)
     nptest.assert_allclose(np.arange(5), matched.matched_data)
 
 
@@ -66,13 +75,16 @@ def test_df_match_match_on_window_border():
     test matching if a value lies exactly on the window border.
     """
 
-    ref_df = pd.DataFrame({"data": np.arange(5)}, index=pd.date_range(datetime(2007, 1, 1, 0),
-                                                                      "2007-01-05", freq="D"))
+    ref_df = pd.DataFrame(
+        {"data": np.arange(5)}, index=pd.date_range(
+            datetime(2007, 1, 1, 0), "2007-01-05", freq="D"))
+
     match_df = pd.DataFrame({"matched_data": np.arange(4)},
                             index=[datetime(2007, 1, 1, 9),
                                    datetime(2007, 1, 2, 9),
                                    datetime(2007, 1, 3, 12),
                                    datetime(2007, 1, 5, 9)])
+
     matched = tmatching.df_match(ref_df, match_df, window=0.5)
 
     nptest.assert_allclose(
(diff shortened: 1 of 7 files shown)


sebhahn:
@cpaulik I fixed the performance issue of df_match and replaced the `groupby` statement. I've also added a new implementation of temporal matching, which has a different interface and return values. However, the current matching function (`df_match`) returns a tuple of dataframes, do you think this is a better idea/format than returning one dataframe with all columns matched. One problem are duplicated column names.

sebhahn:
@cpaulik I think the pull request is almost ready, except for discussing the return options/values of the new temporal matching function. 

I decided not to use the new temporal matching function in `df_match` because it takes quite some effort to achieve the same return format. Now there are two temporal matching functions with two (not too) different interfaces and return formats, which is probably not ideal from a user perspective because it is not clear which one to use. Do you think we should completely drop the new temporal matching in favor of the old (but fast again) temporal matching? Or should we keep both for a certain time and drop `df_match` in a later release? In my opinion, the new implementation is easier to maintain in the long term. 

Furthermore, I noticed that there are duplicated time stamps in the ASCAT time series product (in the latest CDR and also in the test time series used by pytesmo). The temporal matching now checks for duplicates and removes them, which is also shown as a warning to the user.

cpaulik:
There should be one temporal matching function.

So the two functions can be merged if the features of the new function are needed.

- the `asym_window` can be dropped from `df_match`. We should then be able to use `merge_asof` if I understood correctly. Unless this makes the tuple return format too complicated then I wouldn't do it.
- Options like `duplicate_nan` from the new function can easily be added to `df_match`
- I think returning a tuple of dataframes is easier to handle because it guarantees the same order of data as what was given into the `df_match` function. If everything is merged into one dataframe there is the problem of duplicate column names and it is harder to get the data back to the dataset it belongs to.

sebhahn:
@cpaulik I've tried my best to harmonize the temporal matching function into one, but things are now broken in the validation framework. I'm not sure if it's better fixing all those errors or keeping two matching functions?

sebhahn:
I was able to fix all errors except for two. @cpaulik I'm a little bit lost with all this matching in the framework, would it be possible for you to have a look at that?

cpaulik:
I'll try to take a look.

tracyscanlon:
Hi, I'm going to release a new version today without including this pull request as it seems to still have issues. What's the plan on getting this finalised? Thanks, Tracy.

cpaulik:
I still plan to take a look at this. But I've been very busy with other stuff. Sorry for the delay

tracyscanlon:
No worries. Was just wondering. Maybe we can include it in the next release.

tests/test_temporal_matching.py
```diff
@@ -32,67 +31,40 @@
...
                                    datetime(2007, 1, 4, 9),
                                    datetime(2007, 1, 5, 9)])
-    matched = tmatching.df_match(ref_df, match_df)
-
-    nptest.assert_allclose(
-        np.array([0.375, 0.375, 0.375, 0.375, 0.375]), matched.distance.values)
-    nptest.assert_allclose(np.arange(5), matched.matched_data)
-
 
-def test_df_match_match_on_window_border():
```
cpaulik, line 64: This test was removed because of the removal for the `asym_window` options?

Since there is no functional replacement for this option with the new parameters the [cci code](https://git.eodc.eu/cci-sm-work/cci_sm_ecvps_py_src/-/blob/master/smecv/steps/temporal_resampling.py#L85-96) will need to be adapted or we need to add the option back.
@tracyscanlon

wpreimes:
@sebhahn, @cpaulik Reminder that this is still open and should go into the next release, I can have a look at it if no one else has time at the moment

cpaulik:
We need to figure out if we can remove the `asym_windows` option without replacement. See https://github.com/TUW-GEO/pytesmo/pull/176#discussion_r427917250

If you can chase that @wpreimes that would be great of course.

tracyscanlon:
The CCI processor no longer depends on the pytesmo package for temporal resampling. I wrote another implementation using the inbuilt functions in pandas i.e. df.reindex.

wpreimes:
So it should be ok to remove it...I guess I should check @sebhahn code  (👍 for documentation/examples there) with the new `cadati` package in mind and then we should finally replace the old temp. matching.

wpreimes:
@s-scherrer will continue with this and integrate a new version of the temporal matching

s-scherrer:
I compared a minimum implementation of all 3 existing solutions (the one in this PR, the existing in pytesmo, and the one from smecv):

```
temporal_matching_sebastian
21 ms ± 179 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

temporal_matching_pytesmo
63.9 ms ± 3.42 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

temporal_matching_smecv
106 ms ± 735 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Seems like @sebhahn's solution is the fastest of the three, and I will try to make this work. Should the new temporal matching go into `cadati`?

wpreimes:
@tracyscanlon said the CCI implementation is faster than the (original) pytesmo implementation. I also expected the original pytesmo to be much slower. Can you try how it changes if one time series has more time stamps than the other (e.g 1 with hourly data, one with daily)

wpreimes:
Otherwise, good to see that the new version is the fastest, I guess we should continue with this one anyway.

sebhahn:
I think cadati is not the right place, since it is about date/time conversion. It also doesn't depend on pandas at the moment, which I would like to avoid in order to keep it light at this point. I think a better place would be some kind of time series tools package (which doesn't exist at this point) containing functions like: compute_climatology, compute_anomaly, temporal_matching and other common/typical operations applied to geophysical time series

cpaulik:
I'm not up to date a out the other packages you have now. From this uninformed vantage point I would vote for keeping it in pytesmo since the function it replaced is also here.

sebhahn:
Where is the CCI implementation and do all functions return the same results?

s-scherrer:
> Where is the CCI implementation and do all functions return the same results?

The CCI implementation is here (and was the fastest one, see corrected results above):

https://git.eodc.eu/cci-sm-work/cci_sm_ecvps_py_src/-/blob/v06_0/smecv/steps/temporal_resampling.py

The main difference is that it uses `reindex` instead of `merge_asof``. For the comparison, I changed the input and return types so that they are the same, although some of the implementations add some columns to the dataframe. This is one reason for the different runtimes, but the `reindex` is still faster than `merge_asof`.

sebhahn:
Ok, thanks! 

The open question remains if we want/have to re-create the same interface and return values of the new implementation or not?

cpaulik:
I was mainly worried about keeping special features for cci. But since they are no longer necessary I'm happy to deprecate the old function and have a new one with a improved interface.

s-scherrer:
Closed via #204

Issue #165, opened 2019-07-15
Check if metrics.ecol is a real duplicate


cpaulik:
It seems that `metrics.ecol` is in the code two times. Check if the function is really the same and remove one of them

wpreimes:
Duplicate code was removed

PR #179, opened 2019-10-08
Tc metrics update

wpreimes:
Goal was to calculate TC metrics if there are more than 3 temporally matched datasets in one data frame.


wpreimes:
The most important changes are:
- The `nwise_apply()` function in [df_metrics.py](https://github.com/wpreimes/pytesmo/blob/2615d65067e3660510e4620148e716b58f9d12dd/pytesmo/df_metrics.py#L397) now allows applying metrics do more than 2 datasets (before it was called `pairwise_apply()` and that functionality remains with n=2 and as_df=True in the new implementation). Other than that the new function creates a Dictionary for different combinations of columns, and not a matrix as before. This is necessary if `nwise_apply()` takes e.g. 3 datasets.
- Old df metics such as [bias](https://github.com/wpreimes/pytesmo/blob/2615d65067e3660510e4620148e716b58f9d12dd/pytesmo/df_metrics.py#L80) now use the new implementation (I have fixed #138)
- There are `df_metrics` implementations of the [tcol_snr](https://github.com/wpreimes/pytesmo/blob/2615d65067e3660510e4620148e716b58f9d12dd/pytesmo/df_metrics.py#L199) and the [tcol_err](https://github.com/wpreimes/pytesmo/blob/2615d65067e3660510e4620148e716b58f9d12dd/pytesmo/df_metrics.py#L167) metrics that use the `n=3` implementation of `nwise_apply()`. They create a named tuple of the form e.g. `snr(ds1=..,ds2=..,ds3=.. )`
- I have updated the metrics_calculators accordingly
- Everything is tested
- The second ecol function was removed (to close #165)

I am aware that there is the possibility to select triples from more than 3 datasets in the metrics calculator with the (n,k) switch. However, this would not allow to temporally match more than 3 datasets before creating the triples, which is now possible

Should we remove the old [pairwise_apply()](https://github.com/wpreimes/pytesmo/blob/2615d65067e3660510e4620148e716b58f9d12dd/pytesmo/df_metrics.py#L334) function? It is actually not needed anymore now.

Does anyone want to review this? I can explain the changes in more detail if you want.

PR #180, opened 2019-10-31
Add windows release instructions to readme file

cpaulik:
I guess this was not clear at all and thus not done for the last release. 

I've fixed it for `v0.7.1` but with this it is also documented somewhat.


wpreimes:
thanks for the docs update, I will upload the whl files in future as well.

PR #181, opened 2019-11-11
CI: Include builds for Python 3.8

cpaulik:
Had to change a small think in julian date conversion. Since values where returned as arrays when a single float value was the input


cpaulik:
This will only work after cartopy has been rebuilt for Python 3.8 https://github.com/conda-forge/cartopy-feedstock/pull/70

wpreimes:
Thanks Christoph, I will merge this

Issue #167, opened 2019-07-15
Remove grids.py and netcdf.py in grid


sebhahn:

wpreimes:
This was done in #182

PR #182, opened 2019-11-12
Refractor

### LINKED ISSUE #166: Move resample code to repurpose
Move resample code to repurpose

https://github.com/TUW-GEO/pytesmo/blob/master/pytesmo/grid/resample.py

wpreimes:
Remove resample code from pytesmo, add repurpose as dependency for backwards compatibility. 
After this is merged, we can close #166 #167


wpreimes:
Regarding deleting the ``io`` module (#168), should we move that code somewhere else?

wpreimes:
@sebhahn @cpaulik any objections? Otherwise I will merge this and create a new release

environment.yml
```diff
@@ -22,6 +22,8 @@ dependencies:
     - pygeobase
     - ascat
     - ismn
+    - repurpose
```
cpaulik: repurpose is now double

cpaulik: Also add the minimum version number that is necessary. Otherwise older versions will be accepted which will break

cpaulik: Just seen it in requirements. So scratch the last comment

cpaulik:
Thanks for taking care of this. I think you should bump the release to major version 2 since this completely removes some modules

cpaulik:
Or maybe at least to 0.8

wpreimes:
> 
> 
> Or maybe at least to 0.8

Thanks for the feedback, I will do that

PR #186, opened 2020-05-19
Rolmetrics writer

wpreimes:
- Update the return values of rolmetrics calculator to be in line with other metic calcs (each element in the returned array represents 1 location, therefore rolmetrics is an array of (time series) arrays...
- Add writer class that works the same as now for non-time series metrics, that can also store time series metrics (indexed ragged format)
- Update tests, remove travis py2.7


wpreimes:
Not sure about the writer class, I wanted to use the Pointdata class from pynetcf, but didnt manage to keep the results files for non-rolling metrics it in line with the old writer (which is the case for the one I have added now), but I'm not sure how good this works when the file gets larger etc.

pytesmo/validation_framework/results_manager.py
```diff
@@ -1,12 +1,20 @@
...
 
 """
-The results manager stores validation results in netcdf format.
+Module description
 """
+# TODO:
+#   (+) 
+#---------
+# NOTES:
+#   -
```
sebhahn, line 10: this can be removed

pytesmo/validation_framework/results_manager.py
```diff
@@ -1,12 +1,20 @@
 # -*- coding: utf-8 -*-
 
 """
-The results manager stores validation results in netcdf format.
+Module description
```
sebhahn, line 4: description is missing

pytesmo/validation_framework/results_manager.py
```diff
@@ -44,53 +52,299 @@ def build_filename(root, key):
 
     return os.path.join(root, '.'.join([fname, ext]))
 
-def netcdf_results_manager(results, save_path, zlib=True):
+
+class PointDataResults(Dataset):
+
+    def __init__(self, filename, zlib=True, read_only=False):
+
+        self.time_unit = 'days since 1900-01-01 00:00:00'
```
sebhahn, line 60: is 1900-01-01 standard or 1970-01-01?

wpreimes: Was like that in the old writer

wpreimes:
> 
> 
>     * Is the writer class also working for the other (non-rolling) metrics? If not what is the difference?
> 
>     * Why would it be a problem if a files gets larger? Is it always a single result file?

Yes, my main focus for now was that it works for non-rolling metrics the same way as before.
As long as the results are collected before writing to file it should be ok, if the file is opened many times to write e.g results for each point I'm not sure how well it works.

wpreimes:
Tests also pass, except py 3.7

sebhahn:
I've fixed the problem for python 3.7. I think we can merge your code and monitor the performance for large data sets

A final question though, does it make sense to provide different result format depending on the metric calculator? Would this make things easier?

Issue #185, opened 2020-05-15
Validation framework does not compare all the possible combinations


cpaulik:
e..g in the test in https://github.com/TUW-GEO/pytesmo/blob/master/tests/test_validation_framwork/test_validation.py#L416

We have 3 datasets `DS1`, `DS2` and `DS3` which are compared with this config:

```python
    process = Validation(
        datasets, 'DS1',
        temporal_matcher=temporal_matchers.BasicTemporalMatching(
            window=1 / 24.0).combinatory_matcher,
        scaling='lin_cdf_match',
        metrics_calculators={
            (3, 2): metrics_calculators.BasicMetrics(other_name='k1').calc_metrics})
```

But in the results the combination between `DS2` and `DS3` is missing.

@wpreimes @tracyscanlon was this ever an issue in QA4SM?

wpreimes:
Aren't the combinations supposed to be created only between the temporal reference and all other datasets? https://github.com/TUW-GEO/pytesmo/blob/63aa4602a0a69ebb958c8ed6ce87461efc49725b/pytesmo/validation_framework/temporal_matchers.py#L124

wpreimes:
In QA4SM this is actually what we do, so no matching between non-reference datasets.

cpaulik:
I think that depends on the `(n, k)` setting. if you run the framework with `(2, 2)` then you always get 2 datasets temporally matched to each other. But we run it with `(3, 2)` which according to the documentation should temporally match the 3 datasets and then give you all combinations of 2 datasets as the results.

Also in the test we get a dataframe like:

```
n_matched_data
{('DS1', 'DS2', 'DS3'):             (DS1, x)  (DS1, y)  (DS2, x)  (DS2, y)  (DS3, x)  (DS3, y)
2000-01-01         0       0.0       0.0       0.0       0.0       0.0

```

But do do it correctly you should temporally match all three and then calculate the basic metrics on the combinations. Otherwise you give an advantage to the `temporal reference` dataset.

cpaulik:
I guess this bug/misunderstanding is the reason for the metric calculators mixing metrics for 2 datasets and more?

@sebhahn same thing for the HSAF metrics I guess.

wpreimes:
Then I guess the combinations have to be created differntly https://github.com/TUW-GEO/pytesmo/blob/63aa4602a0a69ebb958c8ed6ce87461efc49725b/pytesmo/validation_framework/validation.py#L261-L263
[...]

PR #187, opened 2020-05-20
Val Framework compares dataset matches that do not include the temporal reference

cpaulik:
This changes the behavior of the validation framework.

Before this a validation run with `(n, k)` where `k < n` only gave results for
combinations of datasets which included the temporal reference.

e.g. If 3 dataset (DS1, DS2 and DS3) where given and a metric calculator was run
with `(3, 2)` with the temporal reference DS1. Then only the combinations of
`DS1 - DS2` and `DS1 - DS3` gave results.

This rectifies this behavior to also return results for the combination `DS2 -
DS3` in the above example.

Tests still need to be adapted but I want to discuss this first to see if this
is what people expect.

There are also some edge cases. e.g. If 4 datasets are given but the
configuration of the metric calculator would be `(3, 2)`. I think we should
raise an error in these cases because that would mean that the temporal
reference is not included in all matches.

### DIFF
--- a/src/pytesmo/validation_framework/data_manager.py
+++ b/src/pytesmo/validation_framework/data_manager.py
@@ -387,6 +387,33 @@ def flatten(seq):
     return l
 
 
+def get_result_combinations(ds_dict, n=2):
+    """
+    Get all possible combinations dataset columns
+
+    Parameters
+    ----------
+    ds_dict: dict
+       Dict of lists containing the dataset names as keys and a list of the
+       columns to read from the dataset as values.
+    n: int
+        Number of datasets for combine with each other.
+        If n=2 always two datasets will be combined into one result.
+        If n=3 always three datasets will be combined into one results and so on.
+        n has to be <= the number of total datasets.
+
+    Returns
+    -------
+    results_names : list of tuples
+        Containing all possible combinations of
+        (dataset_x.column, dataset_y.column)
+        for all datasets in ds_dict
+    """
+    combis = []
+    for key in ds_dict:
+        combis.extend(get_result_names(ds_dict, key, n=n))
+    return sorted(list(set(combis)))
+
 def get_result_names(ds_dict, refkey, n=2):
     """
     Return result names based on all possible combinations based on a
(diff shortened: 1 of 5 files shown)


cpaulik:
This relates to https://github.com/TUW-GEO/pytesmo/issues/185

cpaulik:
The results of the current validation framework should not change at all. You will only get more combinations if `n != k` in the `(n, k)` of the metric calculators.

I will update the tests and add some error checks.

wpreimes:
@cpaulik Reminder that this is still open and should go into the next release, I think I can add the missing tests.

cpaulik:
Yes I know. I'll work on it. It was on my list of todos anyway.

wpreimes:
> 
> 
> Yes I know. I'll work on it. It was on my list of todos anyway.

No problem, just let me know if I can help somehow

cpaulik:
Looking in more detail it is a little tricky for some of the tests that are already there. 

At the moment we have [a test](https://github.com/TUW-GEO/pytesmo/blob/fix-comparison-of-other-dataset-combos/tests/test_validation_framwork/test_validation.py#L278) with 3 input datasets but the metric calculators are specified with `(2, 2)`

So this is actually a weird configuration and I'm not 100% sure if it should be allowed.

If we have 3 input datasets and run with `(2, 2)` like that we have the following issue:

- The scaling reference is not included in all the possible temporal matches. This means scaling would fail for some combinations.

So what should happen in that case?

1. Is a warning enough and we just skip the combinations that do not inlcude the scaling reference during the temporal matching?
2. We include the scaling reference in all possible temporal matchings. But that negates the `(2, 2)` setting by including 3 datasets during the temporal matching
3. We raise an error in the beginning and do not allow `(n, k)` with n < total number of datasets?

The original idea was as far as I can remember to allow one validation to run against different datasets independently. e.g. ASCAT as the reference and we validate against ERA and GLDAS independently. It is in that case more efficient to read the ASCAT data only once. 

Can you think of other cases where `n < total number of datasets` makes sense?

cpaulik:
I'm going to implement solution 3. since I can't think of a case where `n` should be smaller than the total number of datasets.

wpreimes:
@cpaulik thanks for finishing this, I will have a look and merge it!

wpreimes (approved):

PR #188, opened 2020-05-20
Remove dos line endings in temporal matching

cpaulik:
those lead to merging conflicts and make it impossible to read the diffs

Just encountered those when looking at the temporal matching


cpaulik:
Just to be clear the code is completely unchanged.

PR #190, opened 2020-06-05
Adding AdvancedMaskingAdapter and test

awst-baum:
This PR adds an AdvancedMaskingAdapter that can be used to mask the data from a reader based on a list of filters. The list must contain 3-tuples of filter parameter: data column to filter on, operator to use and threshold to use. Operator can either be taken from a pre-defined list (same as in MaskingAdapter) or be a user-defined callable.
We'd like to use it in QA4SM to implement complex masking for SMOS data and think that it'll be useful in pytesmo in general.

Example of SMOS filter:

    filtered_reader = AdvancedMaskingAdapter(reader, [
        ('Scene_Flags', smos_exclude_bitmask, 0b00001000), 
        ('Soil_Temperature_Level1', '>=', 273), ])        

(where `smos_exclude_bitmask` is a user-defined function)

### DIFF
--- a/pytesmo/validation_framework/adapters.py
+++ b/pytesmo/validation_framework/adapters.py
@@ -160,6 +160,68 @@ class SelfMaskingAdapter(BasicAdapter):
         data = super(SelfMaskingAdapter, self).read(*args, **kwargs)
         return self.__mask(data)
 
+
+class AdvancedMaskingAdapter(BasicAdapter):
+    """
+    Transform the given (reader) class to return a dataset that is masked based
+    on the given list of filters. A filter is a 3-tuple of column_name, operator, and threshold.
+    This class calls the read_ts or read method of the given reader instance, applies all filters
+    separately, ANDs all filters together, and masks the whole dataframe with the result.
+
+    Parameters
+    ----------
+    cls: object
+        has to have a read_ts or read method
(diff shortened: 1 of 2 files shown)


pytesmo/validation_framework/adapters.py
```diff
@@ -160,6 +160,68 @@ def read(self, *args, **kwargs):
         data = super(SelfMaskingAdapter, self).read(*args, **kwargs)
         return self.__mask(data)
 
+
+class AdvancedMaskingAdapter(BasicAdapter):
```
wpreimes, line 164: This is just a strictly more flexible implementation for a `SelfMaskingAdapter` right? Would it make sense to inherit `SelfMaskingAdapter` from this class to show that it basically does the same but with less options? It just might just confuse people which adapter they should use?

awst-baum: It fills the same evolutionale niche as the SelfMaskingAdapter, yeah. Personally, I wouldn't change the SelfMaskingAdapter at all (since I only left it in for compatability). Well, maybe change it to add a deprecation warning - and then remove it some time in the future.
If we add a deprecation warning and a comment a la 'please use `AdvancedMaskingAdapter` instead', this should help users decide what to use.

PR #191, opened 2020-06-08
Rolmetrics

wpreimes:
- Split tests to avoid segfault
- Add adapter for seasonal/monthly metrics
- Update drop timezone info adapter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,8 @@ install:
   - which python
 
 script:
-  - python setup.py test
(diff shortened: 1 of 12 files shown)


wpreimes:
@sebhahn splitting the tests makes them pass...the problem with this solution is that it looks like it breaks our coverage report

wpreimes:
Fixing the coverage issue was pretty straight forward, still, do you think that splitting the tests is a good enough solution?

PR #205, opened 2020-12-15
added min_obs keyword

s-scherrer:
New min_obs keyword argument for Metrics constructor, set to 100 by default.


wpreimes:
whats the reason to go from 10 to 100?

s-scherrer:
> whats the reason to go from 10 to 100?

I had in mind that we wanted to increase this, but I'll change it back to 10.

wpreimes (approved):

PR #206, opened 2021-01-27
Confidence intervals for metrics

s-scherrer:
Once finished, this PR should add options to calculate confidence intervals for the metrics defined in `metrics.py`.

The following pairwise metrics are available:
- [x] `bias`: Standard CIs for sample mean
- [x] `aad` (average absolute deviation): The absolute differences are not normally distributed, standard CIs for sample means are therefore invalid. Bootstrapping is probably the best solution.
- [x] `mad` (median absolute deviation): Again, standard CIs for sample medians are not applicable, bootstrapping is probably the best solution.
- [x] `rmsd` (root mean square deviation): Standard CIs for sample variance are applicable
- [x] `nrmsd` (normalized root mean square deviation): Standard CIs for sample variance are applicable
- [x] `ubrmsd` (unbiased rmsd): Standard CIs for sample variance are applicable
- [x] `mse`: renamed to `mean_squared_error` (while keeping and deprecating the old implementation), and additionally provided the functions `mse_bias`, `mse_var`, and `mse_corr` for the decomposition.
- [x] `pearsonr`: Renamed to `pearson_r`, which now only returns the correlation coefficient, but not the p value. Old version is still available, but deprecated.
- [x] `spearmanr`: Renamed to `spearman_r`, which now only returns the correlation coefficient, but not the p value. Old version is still available, but deprecated.
- [x] `kendalltau`: Renamed to `kendall_tau`, which now only returns the correlation coefficient, but not the p value. Old version is still available, but deprecated.
- [x] `index_of_agreement`: only bootstrap
- [x] `nash_sutcliffe`: only bootstrap

For the triple collocation metrics, I made the following changes:
- [x] `tcol_error` (triple collocation): only bootstrap available. Deprecated, this is already calculated in `tca_snr`.
- [x] `tcol_snr` (triple collocation): only bootstrap available. I renamed this to `tcol_metrics`, since it's not only calculating SNR. Added a function `tcol_metrics_with_bootstrapped_ci`.
- [ ] `ecol`: I'd rather not touch this


s-scherrer:
Analytical CIs for the pairwise metrics can now be calculated via `with_analytical_ci(metric, x, y)` and bootstrapped CIs via `with_bootstrapped_ci(metric, x, y)`.

I deprecated pairwise metric functions that return more than one value (e.g. `mse`, `pearsonr`, etc.) and introduced new versions of them (`msd`, `pearson_r`) that only return a single metric value.
Also, using the `ddof` keyword argument for the MSD metrics is now deprecated.

I also deprecated `tcol_error` and `tcol_snr`, the latter was simply renamed to `tcol_metrics`, since it is really returning multiple metrics.

setup.cfg
```diff
@@ -29,6 +29,7 @@ package_dir =
     =src
 # DON'T CHANGE THE FOLLOWING LINE! IT WILL BE UPDATED BY PYSCAFFOLD!
 setup_requires = pyscaffold>=3.2a0,<3.3a0
+                 numpy>=1.7.0
```
wpreimes, line 32: will pyscaffold update this line in the future and cause the package to break?

s-scherrer: in pyscaffold 4 this won't work anymore, and this should all be put in pyproject.toml, which we don't have yet. I would suggest to leave it but open a new issue (this line is also more of a fix to #208 and should actually be in a separate PR).

wpreimes (approved):

PR #207, opened 2021-02-02
Cdf matching issue 309

pstradio:
This PR provides an update on the CDF scaling methods in `scaling.py` based on Moesinger et al. (2020). It includes:

- The function `cdf_beta_match`, allowing to scale the source to the reference timeseries using the CDF of the reference and the CDF of the source, fitted on a beta distribution function
- An improved methodology for the scaling of values below the second and above the penultimate percentiles
- An improved methodology to dynamically increase the size of percentile bins in case too few observations are found in a bin, avoiding issues with the overfitting of scaling parameters

The improvements have been made available for the `cdf_match` and `lin_cdf_match` functions as well, although all changes are optional to avoid issues with previous uses of Pytesmo.

### DIFF
--- a/src/pytesmo/scaling.py
+++ b/src/pytesmo/scaling.py
@@ -9,7 +9,7 @@ import numpy as np
 import pandas as pd
 import scipy.interpolate as sc_int
 from warnings import warn
-from pytesmo.utils import unique_percentiles_interpolate
+from pytesmo.utils import unique_percentiles_interpolate, scale_edges
 
 
 def add_scaled(df, method='linreg', label_in=None, label_scale=None):
@@ -279,11 +279,12 @@ def lin_cdf_match(src, ref,
     perc_src = np.array(np.percentile(src, percentiles))
     perc_ref = np.array(np.percentile(ref, percentiles))
 
-    return lin_cdf_match_stored_params(src, perc_src, perc_ref,
+    return lin_cdf_match_stored_params(src, perc_src, perc_ref, ref=ref,
                                        min_val=min_val, max_val=max_val)
 
 
 def lin_cdf_match_stored_params(src, perc_src, perc_ref,
+                                ref=None,
                                 min_val=None, max_val=None):
     """
     Performs cdf matching using given percentiles.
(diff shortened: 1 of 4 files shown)


pstradio:
The last commit adds some tests for the changes that I've made. @wpreimes This PR does not yet include the deprecation of the previous cdf matching-based functions

pstradio:
@wpreimes , I have included the deprecation warnings and should be ready to merge

src/pytesmo/utils.py
```diff
@@ -272,4 +275,134 @@ def array_dropna(*arrs):
...
+    x_hi = src[src >= perc_src[-2]] - perc_src[-2]
+    y_hi = ref[ref >= perc_ref[-2]] - perc_ref[-2]
+    
+    # calculate least squares regression parameters
+    def return_regress(x,y,
+                       where,
+                       perc_src=perc_src,
+                       perc_ref=perc_ref):
+        l = min(len(x),len(y))
+        x, y = x[:l], y[:l]
```
s-scherrer, line 317: Could you rename `l` to `n`? `l` is sometimes hard to distinguish from `1`, depending on the used font. (See also https://www.python.org/dev/peps/pep-0008/#names-to-avoid)

src/pytesmo/scaling.py
```diff
@@ -333,30 +352,104 @@ def cdf_match(src, ref,
...
+    perc_ref = utils.unique_percentiles_interpolate(perc_ref,
                                               percentiles=percentiles)
 
-    return gen_cdf_match(src, perc_src, perc_ref,
+    return gen_cdf_match(src, perc_src, perc_ref, ref=ref,
                          min_val=min_val, max_val=max_val,
-                         k=5)
+                         k=5, **kwargs)
 
+def cdf_beta_match(src, ref,
```
s-scherrer, line 381: Could you add a test for this method? Maybe one test whether the rescaling with minobs works, and one that makes sure the maths work as expected.

s-scherrer (changes requested):

src/pytesmo/scaling.py
s-scherrer: For the second one, something like this should work:

```
from scipy.stats import norm

def test_cdf_beta_match():
    n = 1000
    c = np.linspace(1/n, 1-1/n, n)
    x = norm.ppf(c)
    y = 2 * x + 2
    y_scaled = cdf_beta_match(y, x)
   np.testing.assert_almost_equal(y, y_scaled)
```

pstradio: The tests for all scaling methods are already included

pstradio: Under tests/test_scaling.py::test_scaling_method(method)

pstradio: all scaling methods are tested

s-scherrer: Ah, nice, that's fine, I didn't see this in the diff, sorry.

Issue #214, opened 2021-03-05
Remove dependency of time_series/filters.pyx on deprecated numpy API


s-scherrer:
Currently, when setting `define_macros=[("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")]` in `setup.py` for `pytesmo.time_series.filters` the build fails.

s-scherrer:
Closed via #215

Issue #211, opened 2021-02-12
Create pyproject.toml


s-scherrer:
The old way of configuring setup dependencies in setup.cfg won't work anymore with pyscaffold 4 (to be released, also see https://pyscaffold.org/en/latest/updating.html#updating).

s-scherrer:
Closed via #215

PR #215, opened 2021-03-05
Some smaller fixes

s-scherrer:
- added pyproject.toml (#211)
- removed dependency on deprecated numpy API (#214)
- better calculation of moments in cython metrics
- updated PyScaffold to version 4
- use pytest directly in GH actions
- added pre-commit hooks for running flake8 on diff


docs/conf.py
```diff
@@ -132,19 +143,19 @@
 # keep_warnings = False
 
 
-# -- Options for HTML output ---------------------------------------------------
+# -- Options for HTML output -------------------------------------------------
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'sphinx_rtd_theme'
+html_theme = "alabaster"
```
wpreimes, line 150: I think we use the other template normally, I dont like that one ;)

setup.cfg
```diff
@@ -99,6 +100,13 @@ markers =
     full_framework : marks slow test that use the whole validation framework (deselect with '-m "not full_framework"')
     slow : marks slow tests (deselect with '-m "not slow"')
 testpaths = tests
+# This removes some of the warnings that show up with pytest but are not an issue
+filterwarnings =
```
wpreimes, line 104: :+1:

src/pytesmo/__init__.py
```diff
@@ -19,4 +44,4 @@
 
 testdata_path = os.path.join(src_path, '..', 'tests', 'test-data')
```
wpreimes, line 45: I wonder if there is a better solution, in the tests it is nice to import the testdata path... but maybe this should not be here...

src/pytesmo/time_series/filters.pyx
```diff
@@ -38,25 +38,16 @@ import platform
...
-    bint npy_isnan(double x)
+from cython cimport floating
 
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
 @cython.cdivision(True)
-def exp_filter(np.ndarray[DTYPE_d, ndim=1] in_data,
-               np.ndarray[DTYPE_d, ndim=1] in_jd,
+def exp_filter(floating [:] in_data,
```
wpreimes, line 49: for the exponential filter we actually have another package (pyswi) which is not yet public... we should make pyswi public (currently on gitlab.geo.tuwien.ac.at) and then remove it from pytesmo. @sebhahn and @Adeaem were working on this

.github/workflows/ci.yml
```diff
@@ -54,8 +54,8 @@ jobs:
         shell: bash -l {0}
         run: |
           python setup.py install
-          python setup.py test --addopts "--cache-clear -m 'not full_framework'"
-          python setup.py test --addopts "--cache-clear -m 'full_framework'"
+          pytest --cache-clear -m 'not full_framework'
```
wpreimes, line 57: btw, this was the fix for avoiding the segfault. no idea where they came from...

tests/test_validation_framework/test_adapters.py
```diff
@@ -12,191 +12,251 @@
...
 from tests.test_validation_framework.test_datasets import TestDataset
-from ascat.read_native.cdr import AscatSsmCdr
 
-import numpy as np
-import numpy.testing as nptest
-import pandas as pd
-from datetime import datetime
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore")
+    from ascat.read_native.cdr import AscatSsmCdr
```
wpreimes, line 52: why is the ascat package throwing so many warnings anyway @sebhahn ?

wpreimes (changes requested):
Can you change the docs template please? Otherwise looks fine

src/pytesmo/__init__.py
s-scherrer: I think there is a better solution, but I have to look it up

wpreimes:
I wonder if you need to add `coverage` here https://github.com/s-scherrer/pytesmo/blob/916defdc284288b33fa2fb666d98b85f9408f825/setup.cfg#L65

.coveragerc
```diff
@@ -2,12 +2,6 @@
 [run]
 branch = True
 source = pytesmo
-# omit = bad_file.py
-
-[paths]
-source =
-    src/
-    */site-packages/
```
s-scherrer, line 10: This was causing the issue with coverage, @wpreimes

src/pytesmo/time_series/filters.pyx
s-scherrer: I only updated the existing code here to not depend on the deprecated numpy API anymore. I would suggest to postpone the removal to another PR, so I can merge this here in soon.

wpreimes: sure, we should only remove it when pyswi is public

src/pytesmo/__init__.py
s-scherrer: testdata_path is only used in the examples anyways, I changed the examples accordingly and removed it from the package.

s-scherrer:
- added pyproject.toml (#211)
- removed dependency on deprecated numpy API (#214)
- better calculation of moments in cython metrics
- updated PyScaffold to version 4
- use pytest directly in GH actions
- added pre-commit hooks for running flake8 on diff

PR #217, opened 2021-03-08
Fix bug in indexing when interpolate leapday was set

cpaulik:
If input data is less than 1 year then this failed because it tried to index
days that did not exist

### DIFF
--- a/src/pytesmo/time_series/anomaly.py
+++ b/src/pytesmo/time_series/anomaly.py
@@ -187,11 +187,6 @@ def calc_climatology(Ser,
     clim_ser = pd.Series(clim.values.flatten(),
                          index=clim.index.values)
 
-    if interpolate_leapday and not respect_leap_years:
-        clim_ser[60] = np.mean((clim_ser[59], clim_ser[61]))
-    elif interpolate_leapday and respect_leap_years:
-        clim_ser[366] = np.mean((clim_ser[365], clim_ser[1]))
-
     if wraparound:
         index_old = clim_ser.index.copy()
         left_mirror = clim_ser.iloc[-moving_avg_clim:]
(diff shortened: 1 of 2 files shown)


wpreimes (approved):
Thanks

cpaulik:
@wpreimes Do you have any plans for a next release. I've seen quite a lot of stuff happening lately.

I fixed this hoping we could make a bugfix release quite soon since this is holding us back. If you don't have any problems with it I can also tag a release so it is not work for you.

wpreimes:
The plan was to go for a release by the end of this week. I would like to run a few validations with the updated metrics first. But if it is really urgent I can also do the release now, and if really necessary we could release another fix this week. @s-scherrer your changes are now all in the master, right?

s-scherrer:
Yes, the PRs I opened are all merged. I'm still working on the updated metrics calculators, but this might take a few more days. Releasing now is fine with me.

cpaulik:
I can also put my commit into a hotfix branch and just release 0.9.1. Then we can do a larger release when you are ready.

wpreimes:
> 
> 
> I can also put my commit into a hotfix branch and just release 0.9.1. Then we can do a larger release when you are ready.

ok, I think thats better. 0.10.1 would be the new version

cpaulik:
Yes. just saw 0.10.1.

I can make that on a `v0.10.x` branch

cpaulik:
Pushed the branch, updated the changelog and will tag as v0.10.1

wpreimes:
the `.whl` are included btw: https://github.com/TUW-GEO/pytesmo/actions/runs/632955582#artifacts . Do you have access to pypi still, or should I upload them?

cpaulik:
Yes. I've seen the new CI pipeline using Github Actions. It's nice.

I've uploaded the windows wheels and the tar.gz file since the linux wheels are not compatible with pypi at the moment. If I understood correctly.

wpreimes:
yes. thanks

Issue #218, opened 2021-03-11
dropduplicates=True only drops duplicates in non-ref in df_match


s-scherrer:
When trying to replace `df_match` with the new implementation `temporal_collocation` in `BasicTemporalMatcher` I encountered a strange issue with `dropduplicates`, namely that only duplicates in the non-reference frame will be dropped (see example). Is this the intended behaviour, @wpreimes @sebhahn ?

```
index1 = pd.date_range("2000-01-01", "2000-01-05", freq="D")
index2 = pd.date_range("2000-01-01 03:00", "2000-01-05 03:00", freq="D")
index1.values[2] = index1.values[3]
df1 = pd.DataFrame({"k1": np.arange(len(index1))}, index=index1)
df2 = pd.DataFrame({"k2": np.arange(len(index2))}, index=index2)
```
The dataframes look like this:
```
In [28]: df1
Out[28]: 
            k1
2000-01-01   0
2000-01-02   1
2000-01-04   2
2000-01-04   3
2000-01-05   4

In [29]: df2
Out[29]: 
                     k2
2000-01-01 03:00:00   0
2000-01-02 03:00:00   1
2000-01-03 03:00:00   2
2000-01-04 03:00:00   3
2000-01-05 03:00:00   4
```

When using the frame with duplicates in the index I get the following result:
```
In [32]: df_match(df1, df2, dropduplicates=True)
Out[32]: 
            distance  k2               index
2000-01-01     0.125   0 2000-01-01 03:00:00
2000-01-02     0.125   1 2000-01-02 03:00:00
2000-01-04     0.125   3 2000-01-04 03:00:00
2000-01-04     0.125   3 2000-01-04 03:00:00
2000-01-05     0.125   4 2000-01-05 03:00:00
```
The 2000-01-04 entry is duplicated in the results, which is not really what I would expect from the docstring when using `dropduplicates=True`:
```
    dropduplicates : boolean
        Drop duplicated temporal matched (default: False)
```

sebhahn:
Could you please point/link to the version/commit of df_match you are referring to?

s-scherrer:
I'm referring to the current master branch.

sebhahn:
Did you set `dropduplicates` to True? If it is False nothing is changed in the result

s-scherrer:
Yes:

```
In [32]: df_match(df1, df2, dropduplicates=True)
Out[32]: 
            distance  k2               index
[...]

PR #219, opened 2021-03-12
Improved QA4SM metrics calculators

s-scherrer:
- [x] new matcher that makes sure all dataframe combinations only have common timestamps
- [x] new implementation of QA4SM metrics calculators that take only 2 or 3 dataframes


s-scherrer:
The previous commit compared the old and the new temporal matching implementation when dropping duplicates before matching (see also #218). The CI failed only due to coverage, the tests are running. The next commit will remove the comparison test and update the target values in the tests affected by removing duplicates in the reference.

wpreimes (approved):
@s-scherrer Looks good to me, is this done from your side? Im still a bit worried about #220, but it seems that the current solution is ok for now.

s-scherrer:
I changed the tests a bit to also have a case where the reference name and the other name have a different alphabetical order. The results don't change much, it's just a bit hacky to figure out which one was the reference.

For QA4SM this shouldn't be a problem, because the dataset names are numbered, e.g. `0-ERA5`, `1-ISMN` depending on the order they are given, so they are already alphabetically ordered.

In other cases it's only a usability issue - it's not as easy to tell which one was the reference, but the results are still okay. I would propose to solve #220 in a separate PR.

PR #236, opened 2021-05-17
Data averager issue 106

pstradio:
PR to use averaging/upscaling methods in pytesmo.

- [x] Adapt method for AdvancedMaskingAdapter class
- [x] Create tests
- [x] Setup validations configuration with non-reference ISMN


src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+    @property
+    def lut(self) -> dict:
+        """Get a lookup table that combines the points falling under the same reference pixel"""
+        lut = {}
+        for other_name, other_class in self.others_class.items():
+            try:
+                grid = other_class.grid
+            except AttributeError:
+                # when filters are applied, this will be an AdvancedMaskingAdapter
+                grid = other_class.cls.cls.grid
```
s-scherrer, line 159: Use `hasattr(other_class, "grid")` instead of the `try ... except` approach. I think you should also not restrict this to only two adapters, but use a loop to check whether it has "grid" or "cls" as attributes.

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+                both=True,
+            )
+            other_lut = {}
+            # iterate from the side of the non-reference
+            for gpi, lon, lat in zip(other_points[0], other_points[1], other_points[2]):
+                # list all non-ref points under the same ref gpi
+                try:
+                    ref_gpi = self.ref_class.grid.find_nearest_gpi(lon, lat)[0]
+                except AttributeError:
+                    ref_gpi = self.ref_class.cls.cls.grid.find_nearest_gpi(lon, lat)[0]
```
s-scherrer, line 172: Same as above. In general, if you can figure out stuff by a simple check like `hasattr`, always do this instead of `try - except`.

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+        dss: list
+            list of dataframes of the reference timeseries
+        """
+        dss = []
+        for gpi, lon, lat in points:
+            # todo: resolve ambiguity on gpi/lon, lat and AssertionError on validator/hacks
+            try:
+                ds = self.read_ds(other_name, gpi)
+            except AttributeError:
+                continue
```
s-scherrer: Again, use `hasattr`

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+            dss.append(ds)
+
+        return dss
+
+    @staticmethod
+    def temp_match(
```
s-scherrer, line 210: Why are you not using the temporal matching module here?

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+            )
+            return df
+
+        return filtered
+
+    def upscale(self, df, method='average', **kwargs) -> pd.DataFrame:
+        """
+        Return the upscaled Dataframe with the specified method
+
+        **New upscaling methods can be specified here in the lut**
```
s-scherrer, line 322: Move this comment outside of the docstring, it's only relevant for developers. I would just put it in the line above `up_function`. Also, I'm not sure how universally "lut" is understood as look-up table, took me some time to figure it out, so maybe just write it out in the comment.

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+            dataframe with "upscaled" column
+        """
+        up_function = {
+            "average": self.faverage,
+        }
+        f = up_function[method]
+        # check that column names are all the same and return first
+        g = groupby(df.columns)
+        assert (next(g, True) and not next(g, False))
+        var_name = df.columns[0]
```
s-scherrer: An alternative to check that column names are all the same would be `assert len(set(df.columns)) == 1`, which I would consider more pythonic ;)

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+        out = df.mean(axis=1)
+
+        return out
+
+    def wrapper(
+            self,
+            gpi,
+            other_name,
+            tmatching="rescale",
+            up_method="average",
```
s-scherrer, line 358: Can you call this "upscaling_method"? It's a bit more verbose. Probably in the line above "temporal_matching_method" would also be better.

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+
+    @staticmethod
+    def faverage(df) -> pd.Series:
+        """Simple average of each column in the dataframe"""
+        # workaround to avoid changes in place
+        out = df.mean(axis=1)
+
+        return out
+
+    def wrapper(
```
s-scherrer, line 353: "wrapper" is not a very descriptive function name in this case.

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+
+        upscaled = pd.DataFrame(
+            data=f(df, **kwargs),
+            columns=[var_name]
+        )
+
+        return upscaled
+
+    @staticmethod
+    def faverage(df) -> pd.Series:
```
s-scherrer, line 346: Functions that are only intended for internal use should start with an underscore, I would rename this to `_average`.

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+                * 'average' takes the simple mean of all timeseries
+        tstability: bool, default is False
+            if True, the values are filtered using the time stability concept
+        kwargs: keyword arguments
+            argumjents for the temporal window or time stability thresholds
+
+        Returns
+        -------
+        upscaled: pd.DataFrame
+            upscaled time series
```
s-scherrer, line 385: Document that `None` is also a possible return value

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+
+    def wrapper(
+            self,
+            gpi,
+            other_name,
+            tmatching="rescale",
+            up_method="average",
+            tstability=False,
+            **kwargs
+    ) -> pd.DataFrame:
```
s-scherrer, line 361: With `None` as possible return value, the type hint should be `Union[None, pd.DataFrame]` (with `from typing import Union`)

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,423 @@
...
+        # check that timeseries were extracted from the points
+        if not tss:
+            return None
+        tss = [df for df in tss if not df is None]
+
+        # handle situation with single timeseries or all None
+        if len(tss) < 1:
+            if not tss:
+                return None
+            return tss[0]
```
s-scherrer: This should be:
```
if not tss:
    return None
```
The second return statement does not make sens, since `len(tss) < 1`, so `tss[0]` does not exist.

src/pytesmo/validation_framework/data_manager.py
```diff
@@ -90,15 +94,19 @@ class DataManager(object):
...
-        """
+    def __init__(
+            self, datasets,
+            ref_name,
+            period=None,
+            read_ts_names='read_ts',
+            upscale_parms=None,
+            geo_subset=None,
+    ):
+        """Initialize parameters."""
```
s-scherrer, line 105: I know, it was already here, but this comment is useless, you can remove it.

s-scherrer (changes requested):
Looks good so far, only a few technical issues.

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,430 @@
...
+
+        Returns
+        -------
+        matched: pd.DataFrame
+            dataframe with temporally matched timeseries
+        """
+        if method == 'common':
+            matched = pd.concat(to_match, axis=1, join="inner").dropna(how="any")
+
+        elif method == 'rescale':
```
s-scherrer, line 238: I think 'rescale' is a bit confusing here, because in pytesmo terminology, rescaling is the stuff that's done in the scaling module (CDF matching, min-max-scaling, mean-std-scaling

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,430 @@
...
+                points = len(df.dropna())
+                if n == 0:
+                    ref = df
+                if len(ref.dropna()) >= points:
+                    continue
+                else:
+                    ref = df
+
+            to_match = pd.concat(to_match, axis=1)
+            matched = tcoll(
```
s-scherrer, line 252: Can you use the full name, i.e `temporal_collocation` instead of `tcoll`?

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,430 @@
...
+                ref,
+                to_match,
+                pd.Timedelta(hours, "H"),
+                dropna=True
```
s-scherrer, line 256: If you are calling `dropna(how="any")` on it afterwards, you can set this to False here.

src/pytesmo/validation_framework/data_averaging.py
s-scherrer: Ah, right, I didn't recognize it due to the different name.

src/pytesmo/validation_framework/data_averaging.py
pstradio: That's true. anyways, I have decided that it doesn't make sense to have both methods and I'll only keep the temporal matching option

src/pytesmo/validation_framework/data_averaging.py
pstradio: Not exactly, because the function `temporal_collocation` uses `dropna(how="all")` but we may want to drop all entries if only one point lacks values

s-scherrer: I meant you can set `dropna=False` in the call to `temporal_collocation`, i.e.:
```
matched = temporal_collocation(..., dropna=False).dropna(how="any")
```

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,428 @@
...
+                dropna=True,
+                checkna=True,
+            )
+            matched.append(matched_ser)
+        matched = pd.concat(matched, axis=1)
+
+        if drop_missing:
+            matched.dropna(axis=0, how="any", inplace=True)
+
+        return matched
```
s-scherrer, line 259: There's also `combined_temporal_collocation`, which should probably already do what you want (and a bit more).

https://github.com/TUW-GEO/pytesmo/blob/5e4a54662b935402019b0c232eaae7b79b284494/src/pytesmo/temporal_matching.py#L400

src/pytesmo/validation_framework/data_averaging.py
```diff
@@ -0,0 +1,428 @@
...
+
+    @property
+    def lut(self) -> dict:
+        """Get a lookup table that combines the points falling under the same reference pixel"""
+        lut = {}
+        for other_name, other_class in self.others_class.items():
+            if hasattr(other_class, "grid"):  # todo: check potential object attributes
+                grid = other_class.grid
+            else:
+                grid = other_class.cls.cls.grid
```
s-scherrer, line 158: I think we should fix the adapters instead of relying on this approach. You can leave it for now, but let's add a task for this sprint.

s-scherrer: I actually already fixed it. If you rebase/merge the current master, `grid = other_class.grid` should also work with adapters (but the grid could be None in case the original reader had no grid).

s-scherrer (approved):

src/pytesmo/validation_framework/upscaling.py
```diff
@@ -0,0 +1,391 @@
...
+    The goal is to include here all identified upscaling methods to provide an estimate at the reference footprint scale.
+
+    Implemented methods:
+
+        * time-stability filtering
+        * simple averaging
+
+    Parameters
+    ----------
+    ref_class: pygeogrids Grid obj.
```
s-scherrer, line 125: There should be a space also before the colon, otherwise the documentation won't render correctly. I think you need to change this in all docstrings.

src/pytesmo/validation_framework/upscaling.py
```diff
@@ -0,0 +1,391 @@
...
+            gpi value of the reference point
+        other_name: str
+            name of the non-reference dataset to be upscaled
+        upscaling_method: str
+            method to use for upscaling:
+                * 'average' takes the simple mean of all timeseries
+        temporal_stability: bool, default is False
+            if True, the values are filtered using the time stability concept
+        kwargs: keyword arguments
+            argumjents for the temporal window or time stability thresholds
```
s-scherrer, line 340: typo

src/pytesmo/validation_framework/upscaling.py
```diff
@@ -0,0 +1,391 @@
...
+            point_ts = point_df[target_column]
+            to_match.append(
+                point_ts.to_frame(name=target_column + "_{}".format(n))  # avoid name clashing
+            )
+
+        try:
+            tss = self.temporal_match(to_match, **kwargs)
+        # in case no match is found (result is all nans)
+        except UserWarning:
+            return None
```
s-scherrer, line 380: Did you test if catching a warning like this works? I thought you need to use the `warnings` module for stuff like this.

tests/test_validation_framework/test_upscaling.py
```diff
@@ -0,0 +1,120 @@
...
+import warnings
+
+import pandas as pd
+import numpy as np
+from datetime import datetime
+
+from pytesmo.validation_framework.upscaling import Upscaling
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore")
```
s-scherrer, line 40: If I understand the `warnings` module correctly, this statement has no effect, since the filter is only applied inside the `with` statement.

tests/test_validation_framework/test_upscaling.py
```diff
@@ -0,0 +1,120 @@
...
+        "read_ts_names": None,
+    }
+    averager = Upscaling(
+        ref_class=None,
+        others_class={},
+        upscaling_lut={},
+        manager_parms=manager_parms
+    )
+
+    return averager
```
s-scherrer, line 56: You could make this a pytest.fixture:
```
@pytest.fixture
def averager():
    ...
    return averager
```

and then use it as function argument for the tests:

```
def test_upscale(averager):
    ...
    upscaled = averager.upscale(to_upscale, method="average")
    ...
```

tests/test_validation_framework/test_upscaling.py
```diff
@@ -0,0 +1,120 @@
...
+def test_temporal_matching():
+    """Test temporal matching"""
+    averager = return_averager()
+    data_ref = np.arange(30.)
+    data2match = data_ref[:-1]
+    data2match[2] = np.nan
+
+    ref_ser = pd.Series(
+        data_ref,
+        index=pd.date_range(datetime(2007, 1, 1, 0), datetime(2007, 1, 30, 0), freq="D"),
```
s-scherrer, line 105: Just FYI: You can also write `pd.date_range("2007-01-01", "2007-01-30", freq="D")` instead. Is a bit less typing and you don't need to import datetime and wonder whether it's datetime.datetime or only datetime ;)

tests/test_validation_framework/test_validation.py
```diff
@@ -426,6 +426,146 @@ def test_ascat_ismn_validation_metadata(ascat_reader):
...
+    # Initialize ISMN reader
+    ismn_data_folder = os.path.join(
+        os.path.dirname(__file__),
+        "..",
+        "test-data",
+        "ismn",
+        "multinetwork",
+        "header_values",
+    )
+    ismn_reader = ISMN_Interface(ismn_data_folder)
```
s-scherrer, line 461: Not exactly part of your PR, but while you're at it, could you make this a fixture like `ascat_reader`? Then it's easier to reuse it for other tests.

tests/test_validation_framework/test_validation.py
```diff
@@ -426,6 +426,146 @@ def test_ascat_ismn_validation_metadata(ascat_reader):
...
+                "mask_frozen_prob": 80,
+                "mask_snow_prob": 80,
+                "mask_ssf": True,
+            }
+        },
+        "ISMN": {
+            "class": ismn_reader,
+            "columns": ["soil moisture"],
+        },
+    }
```
s-scherrer, line 486: This is also used exactly like this in a lot of tests, could you make this a fixture too?

tests/test_validation_framework/test_validation.py
```diff
@@ -426,6 +426,146 @@ def test_ascat_ismn_validation_metadata(ascat_reader):
...
+    r_should = np.array(
+        [-0.012335,  0.257671,  0.657239],
+        dtype=np.float32,
+    )
+
+    assert sorted(vars_should) == sorted(calc_vars)
+    assert n_obs == n_obs_should
+    nptest.assert_allclose(sorted(rho), sorted(rho_should), rtol=1e-4)
+    nptest.assert_allclose(sorted(rmsd), sorted(rmsd_should), rtol=1e-4)
+    nptest.assert_allclose(sorted(r), sorted(r_should), rtol=1e-4)
```
s-scherrer, line 567: This is the same as in other tests, could you make a function out of this instead?

tests/test_validation_framework/test_upscaling.py
```diff
@@ -0,0 +1,120 @@
...
+    to_match = [ref_ser, match_ser]
+
+    matched = averager.temporal_match(to_match, drop_missing=False)
+    assert len(matched.index) == 30, "Should be matched to the longest timeseries"
+
+    matched = averager.temporal_match(to_match, drop_missing=True)
+    assert len(matched.index) == 28, "Should drop the row and the missing timestep with a missing value"
+
+    matched = averager.temporal_match(to_match, hours=4)
+    assert matched[matched.columns[1]].dropna().empty, "Should not be matched"
```
s-scherrer, line 120: Can you also add a test for `get_upscaled_ts` that triggers the warning that you are trying to catch?

s-scherrer (changes requested):
Good PR, thanks :)

Mostly minor issues, only a missing test.

src/pytesmo/validation_framework/upscaling.py
```diff
@@ -0,0 +1,391 @@
...
+
+    Implemented methods:
+
+        * time-stability filtering
+        * simple averaging
+
+    Parameters
+    ----------
+    ref_class: pygeogrids Grid obj.
+        Class containing the method read_ts for reading the data of the reference
```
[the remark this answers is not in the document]
pstradio: You are right, it it actually a reader class

src/pytesmo/validation_framework/upscaling.py
```diff
@@ -0,0 +1,391 @@
...
+    ):
+        self.ref_class = ref_class
+        self.others_class = others_class
+        self.lut = upscaling_lut
+        # attributes used by the Mixin class:
+        self.datasets = manager_parms["datasets"]
+        self.period = manager_parms["period"]
+        self.read_ts_names = manager_parms["read_ts_names"]
+
+    def get_timeseries(
```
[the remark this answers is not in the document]
pstradio: I changed it to `_read`; it is really a method that is used inside the class only but I agree this way it's more consistent

tests/test_validation_framework/test_validation.py
pstradio: I have created this function, but the values need to be specified test-by-test regardless :/ ; still, a bit cleaner in the code

tests/test_validation_framework/test_upscaling.py
pstradio: I have decided that instead of returning None it is better to just drop the series anyway.

src/pytesmo/validation_framework/upscaling.py
```diff
@@ -221,6 +221,7 @@ def temporal_match(
             combined_dropna=combined_dropna,
             checkna=True,
         )
+        matched.dropna(axis="columns", how="all", inplace=True)
```
s-scherrer, line 224: Did `combined_dropna` not work as expected?

pstradio: The thing is that it only works on rows but this is to eliminate a whole column

PR #241, opened 2021-06-06
Use assert_almost_equal in deprecated MSE test

cpaulik:
This failed for on a machine (osx arm64) with

```
    def test_mse(arange_testdata):
        """
        Test for mse
        """
        with pytest.deprecated_call():
            # example 1
            x, y = arange_testdata

            mse_pred = 4.
            mse_bias_pred = 2. ** 2
            mse_obs, _, mse_bias, _ = met.mse(x, y)

>           nptest.assert_equal(mse_obs, mse_pred)
E           AssertionError:
E           Items are not equal:
E            ACTUAL: 4.0000000000000036
E            DESIRED: 4.0
```

### DIFF
--- a/tests/test_metrics_deprecated.py
+++ b/tests/test_metrics_deprecated.py
@@ -143,8 +143,8 @@ def test_mse(arange_testdata):
         mse_bias_pred = 2. ** 2
         mse_obs, _, mse_bias, _ = met.mse(x, y)
 
-        nptest.assert_equal(mse_obs, mse_pred)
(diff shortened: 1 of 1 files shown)


wpreimes:
thanks, Is this because of the new hardware architecture? I wonder if there will be more cases like this...

cpaulik:
Yes, I guess it is most likely because of floating point calculation differences between Intel and ARM

For pytesmo I ran the whole testsuite and this was the only one that was failing.

PR #243, opened 2021-06-22
Add adapter to combine multiple columns returned by read function

wpreimes:
Recently I had the case where we had multiple layers of RZSM in a dataset and wanted to compare it to a model which had one (much wider) layer as the reference. So I wrote this adapter to merge columns upon passing them into the validation framework. What do you think @s-scherrer @pstradio does this make sense here?

### DIFF
--- a/src/pytesmo/validation_framework/adapters.py
+++ b/src/pytesmo/validation_framework/adapters.py
@@ -36,6 +36,7 @@ import operator
 from pytesmo.time_series.anomaly import calc_anomaly
 from pytesmo.time_series.anomaly import calc_climatology
 from pandas import DataFrame
+import pandas as pd
 
 
 class BasicAdapter(object):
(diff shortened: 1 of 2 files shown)


src/pytesmo/validation_framework/adapters.py
```diff
@@ -314,3 +314,51 @@ def read_ts(self, *args, **kwargs):
...
+        data[self.new_name] = new_col
+        return data
+
+    def read_ts(self, *args, **kwargs) -> DataFrame:
+        data = super(ColumnCombAdapter, self).read_ts(*args, **kwargs)
+        return self.apply(data)
+
+    def read(self, *args, **kwargs) -> DataFrame:
+        data = super(ColumnCombAdapter, self).read(*args, **kwargs)
+        return self.apply(data)
```
s-scherrer, line 364: Can you change this to
```
return self.read_ts(*args, **kwargs)
```
or change `read_ts` to call `read`?

src/pytesmo/validation_framework/adapters.py
```diff
@@ -314,3 +314,51 @@ def read_ts(self, *args, **kwargs):
...
+        cls : class instance
+            Must have a read_ts or read method returning a pandas.DataFrame
+        func: Callable
+            Will be applied to dataframe columns using pd.DataFrame.apply(..., axis=1)
+            additional kwargs for this must be given in func_kwargs,
+            e.g. pd.DataFrame.mean
+        func_kwargs : dict, optional (default: None)
+            kwargs that are passed to method
+        columns: list, optional
+            columns in the dataset that are combined.
```
s-scherrer, line 340: Can you add in the description that in case `columns=None`, all columns are used?

tests/test_validation_framework/test_adapters.py
```diff
@@ -264,3 +265,17 @@ def test_timezone_removal():
...
+    ds = TestDataset("", n=20)
+    orig = ds.read()
+    ds_adapted = ColumnCombAdapter(ds, func=pd.DataFrame.mean, columns=["x", "y"],
+                                func_kwargs={'skipna': True}, new_name='xy_mean')
+    ds_mean1 = ds_adapted.read_ts()
+    ds_mean2 = ds_adapted.read()
+
+    for ds_mean in [ds_mean1, ds_mean2]:
+        nptest.assert_almost_equal(ds_mean["x"].values, orig["x"].values)
+        nptest.assert_almost_equal(ds_mean["y"].values, orig["y"].values)
```
s-scherrer, line 280: These shouldn't change at all, right? So can you use `nptest.assert_equal` instead of `nptest.assert_almost_equal`?

s-scherrer (changes requested):

wpreimes:
Thanks  @s-scherrer I think I changed everything

s-scherrer:
looks good, I'm only a bit unsure whether the last test with `assert_equal` works, there might be some rounding issues. I was only referring to the two lines above, with the columns that shouldn't change at all. But if it works, you can merge it.

s-scherrer (approved):

PR #244, opened 2021-07-02
Fix rollmetric

wpreimes:
Some readers return data not as float64 (but as float32) in that case the rollmetrics may not get data in the expected format. This forces float64 to fix this.

### DIFF
--- a/src/pytesmo/metrics/_fast_pairwise.pyx
+++ b/src/pytesmo/metrics/_fast_pairwise.pyx
@@ -338,10 +338,10 @@ cpdef _pearsonr_from_moments(floating varx, floating vary, floating cov, int n):
 # This implementation is much faster than the old version with numba:
 # old: 76.7 ms ± 1.62 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
 # new: 117 µs ± 836 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
-cpdef rolling_pr_rmsd(floating [:] timestamps,
+cpdef rolling_pr_rmsd(double [:] timestamps,
                       floating [:] x,
                       floating [:] y,
-                      floating window_size,
+                      double window_size,
                       int center,
                       int min_periods):
     """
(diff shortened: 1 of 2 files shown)


s-scherrer:
What's the actual problem with the unexpected data format? Does it fail? 

Maybe it's better to transform everything to the data type that the data got instead of float64, or change the cython code to always expect the times to be doubles (which I think is always the case with pandas to_julian_date).

s-scherrer:
What's the actual problem with the unexpected data format? Does it fail? 

Maybe it's better to transform everything to the data type that the data got instead of float64, or change the cython code to always expect the times to be doubles (which I think is always the case with pandas to_julian_date).

wpreimes:
yes it fails with an Error `Error Buffer dtype mismatch, expected 'double' but got 'float'`
I had 2 situations where one works and one fails with the current implementation

- Rollmetrics: ASCAT SSM to GLDAS as reference: [This](https://github.com/TUW-GEO/pytesmo/blob/master/src/pytesmo/validation_framework/metric_calculators.py#L1261) works, both `xy` and `time` are of type `dtype('float64')`

- Rollmetrics: ASCAT SSM to ERA5-Land as reference: [This](https://github.com/TUW-GEO/pytesmo/blob/master/src/pytesmo/validation_framework/metric_calculators.py#L1261) fails with the error below `time` is still `float64` but `xy` is `dtype('float32')` now. I could fix this case by setting `timestamps.astype('float32')` but this would mean that case 1 fails then. Only the current fix always works and is in line with the docstrings.
```
Traceback (most recent call last):
  File "/opt/pycharm-2021.1/plugins/python/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<string>", line 1, in <module>
  File "src/pytesmo/metrics/_fast_pairwise.pyx", line 342, in pytesmo.metrics._fast_pairwise.__pyx_fuse_1rolling_pr_rmsd
ValueError: Buffer dtype mismatch, expected 'double' but got 'float'
```

wpreimes:
So the problem is not time but xy

s-scherrer:
I think the actual issue is with the cython code:
https://github.com/TUW-GEO/pytesmo/blob/1b543702fde3b6f2228416140edf3896977b9c46/src/pytesmo/metrics/_fast_pairwise.pyx#L341-L344

`timestamps` and `window_size` should be `double`, not `floating`.

wpreimes:
Ok, that seems to fix it

s-scherrer (approved):

PR #245, opened 2021-07-29
Add option for custom reading method name to adapters

wpreimes:
Most changes here (also in C files) are due to applying pre-commit hooks.

The 2 relevant parts are
- In src/pytesmo/validation_framework/adapters.py I added a keyword that allows adding a method other than ``read``/``read_ts`` to the adapters. Data from this method (which is expected to return a DataFrame) is modified in the same way as for read/read_ts. By default, nothing changes compared to the previous implementation.
- In some places I replaced the default `read_ts` function with `read`, as ``read_ts`` is [deprecated in pygeobase](https://github.com/TUW-GEO/pygeobase/blob/master/src/pygeobase/io_base.py#L689) and leads to many warnings when used in the validation framework.

### DIFF
--- a/src/pytesmo/validation_framework/adapters.py
+++ b/src/pytesmo/validation_framework/adapters.py
@@ -1,7 +1,6 @@
 # Copyright (c) 2016,Vienna University of Technology,
 # Department of Geodesy and Geoinformation
 # All rights reserved.
-import warnings
 
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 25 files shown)


src/pytesmo/validation_framework/adapters.py
```diff
@@ -89,28 +134,43 @@ class MaskingAdapter(BasicAdapter):
...
-                          '>=': operator.ge,
-                          '>': operator.gt}
-        self.operator = self.op_lookup[op]
         self.threshold = threshold
 
         self.column_name = column_name
 
+        if self.read_name:
+            setattr(self, self.read_name, self._adapt_custom)
+
```
s-scherrer, line 173: Is this still necessary? This should already happen in the parent class constructor, I think you can remove it.

src/pytesmo/validation_framework/adapters.py
```diff
@@ -89,28 +134,43 @@ class MaskingAdapter(BasicAdapter):
...
+    op: str or Callable
+        one of '<', '<=', '==', '>=', '>', '!=' or a function that takes
+        data and threshold as arguments.
     threshold:
         value to use as the threshold combined with the operator
-    column_name: string, optional
+    column_name: str, optional
         name of the column to cut the read masking dataset to
+    kwargs:
+        Additional kwargs are passed to BasicAdapter.
```
s-scherrer, line 145: Can you add the `read_name` description here and in all other child classes?

src/pytesmo/validation_framework/adapters.py
```diff
@@ -89,28 +134,43 @@ class MaskingAdapter(BasicAdapter):
...
+        super(MaskingAdapter, self).__init__(cls, **kwargs)
+
+        self.op_lookup = {
+            "<": operator.lt,
+            "<=": operator.le,
+            "==": operator.eq,
+            "!=": operator.ne,
+            ">=": operator.ge,
+            ">": operator.gt,
+        }
```
s-scherrer, line 158: Not directly related to your changes, but why is this an attribute? Since this is also used in other adapters, I think it's better to just make this a global variable within this module, maybe something like `_op_lookup`.

src/pytesmo/validation_framework/adapters.py
```diff
@@ -124,6 +184,11 @@ def read(self, *args, **kwargs):
...
         return self.__mask(data)
 
+    def _adapt_custom(self, *args, **kwargs):
+        data = super()._adapt_custom(*args, **kwargs)
+        return self.__mask(data)
+
+
 class SelfMaskingAdapter(BasicAdapter):
     """
     Transform the given (reader) class to return a dataset that is masked based
```
s-scherrer, line 194: What is actually the difference between `MaskingAdapter` and `SelfMaskingAdapter`? They look very redundant to me.

Also, I think those should only override the `_adapt` method of the parent class, the other methods can be removed. Something like:
```
def _adapt(self, data):
    mask = self.operator(data[self.column_name], self.threshold)
    return data[mask]
```

src/pytesmo/validation_framework/adapters.py
```diff
@@ -225,53 +316,70 @@ def __mask(self, data):
...
             ite = data
         else:
             ite = self.columns
         for column in ite:
-            data[column] = calc_anomaly(data[column],
-                                        window_size=self.window_size)
+            data[column] = calc_anomaly(
+                data[column], window_size=self.window_size
+            )
         return data
```
s-scherrer, line 371: Rename to `_adapt` and remove the other methods

src/pytesmo/validation_framework/adapters.py
```diff
@@ -135,76 +200,102 @@ class SelfMaskingAdapter(BasicAdapter):
...
+            ">=": operator.ge,
+            ">": operator.gt,
+        }
 
         self.filter_list = filter_list
 
+        if self.read_name:
+            setattr(self, self.read_name, self._adapt_custom)
+
     def __mask(self, data):
```
s-scherrer, line 299: Rename to `_adapt` and remove the other methods.

src/pytesmo/validation_framework/adapters.py
```diff
@@ -281,22 +389,40 @@ class AnomalyClimAdapter(BasicAdapter):
...
+            cls_kwargs["read_name"] = kwargs.pop("read_name")
+
+        super(AnomalyClimAdapter, self).__init__(cls, **cls_kwargs)
+
         self.kwargs = kwargs
         self.columns = columns
 
+        if self.read_name:
+            setattr(self, self.read_name, self._adapt_custom)
+
```
s-scherrer, line 425: Already in parent constructor

src/pytesmo/validation_framework/adapters.py
```diff
@@ -225,53 +316,70 @@ def __mask(self, data):
...
-        data = super(AdvancedMaskingAdapter, self).read(*args, **kwargs)
+        data = super().read(*args, **kwargs)
         return self.__mask(data)
 
+    def _adapt_custom(self, *args, **kwargs):
+        data = super()._adapt_custom(*args, **kwargs)
+        return self.__mask(data)
+
+
 class AnomalyAdapter(BasicAdapter):
```
s-scherrer, line 331: Would it make sense to merge `AnomalyAdapter` and `AnomalyClimAdapter` to a single class and decide on the different behaviours based on a keyword argument? They are very similar.

src/pytesmo/validation_framework/adapters.py
```diff
@@ -341,24 +482,34 @@ def __init__(self, cls, func, func_kwargs=None, columns=None, new_name='merged')
...
         self.new_name = new_name
 
+        if self.read_name:
+            setattr(self, self.read_name, self._adapt_custom)
+
     def apply(self, data: DataFrame) -> DataFrame:
         columns = data.columns if self.columns is None else self.columns
         new_col = data[columns].apply(self.func, **self.func_kwargs)
         data[self.new_name] = new_col
         return data
```
s-scherrer, line 503: Rename to `_adapt`

src/pytesmo/validation_framework/adapters.py
```diff
@@ -341,24 +482,34 @@ def __init__(self, cls, func, func_kwargs=None, columns=None, new_name='merged')
...
+        super(ColumnCombineAdapter, self).__init__(cls, **kwargs)
         self.func = func
         self.func_kwargs = func_kwargs if func_kwargs is not None else {}
-        self.func_kwargs['axis'] = 1
+        self.func_kwargs["axis"] = 1
         self.columns = columns
         self.new_name = new_name
 
+        if self.read_name:
+            setattr(self, self.read_name, self._adapt_custom)
```
s-scherrer, line 497: Already in parent constructor.

s-scherrer (changes requested):

src/pytesmo/validation_framework/adapters.py
wpreimes: good question, looks like one picks a single column to mask, the other masks the whole data frame...anyway I dont see the point of `MaskingAdapter` (and we dont use that one anywhere as far as I am aware), I would mark it as deprecated and remove it next time.

I think `__adapt` should work (otherwise the parent/child class get mixed up)

s-scherrer: > I think __adapt should work (otherwise the parent/child methods get mixed up)

The parent method can still be called with `super()._adapt`, for example like this:
````
def _adapt(self, data):
    data = super()._adapt(data)
    mask = self.operator(data[self.column_name], self.threshold)
    return data[mask]
```
Do we actually need to distinguish between the methods of parent and child class?

wpreimes: We just have to make sure that all of them are called ;) and if you adapt and already adapted reader as well etc.

src/pytesmo/validation_framework/adapters.py
wpreimes: maybe, they require different arguments (`window_size` only vs many options for the climatology calculation). We can combine them for sure, but as we use them both in qa4sm I would do this in a separate PR.

src/pytesmo/validation_framework/adapters.py
wpreimes: See https://github.com/TUW-GEO/pytesmo/pull/245/commits/602b5b6281a82d151f38a9955ddab67acb877dac
I guess this is what you meant? I like that solution, the only "problem" is that the adaptation function of the BasicAdapter is called multiple time when stacking multiple adapters. This is not really a problem because in those cases nothing happens then. Just a few unnecessary attribute checks.

src/pytesmo/validation_framework/adapters.py
s-scherrer: Yes, probably better

s-scherrer (approved):

PR #249, opened 2021-10-15
Monthly climatology

pstradio:
Simple method adaptation to compute a month-based climatology in `anomalies.calc_climatology`

- [x] Update test
- [x] Update `calc_anomalies` to calculate monthly anomalies (?)

### DIFF
--- a/src/pytesmo/time_series/anomaly.py
+++ b/src/pytesmo/time_series/anomaly.py
@@ -168,7 +168,8 @@ def calc_climatology(Ser,
         moving average applied on the calculated climatology
 
     unit: str, optional
-        Unit of the year to apply the climatology calculation to. Currently supported options are 'day', 'month'.
+        Unit of the year to apply the climatology calculation to. Currently,
+        supported options are 'day', 'month'.
         Default: 'day'
 
     Returns
--- a/src/pytesmo/time_series/anomaly.py
+++ b/src/pytesmo/time_series/anomaly.py
@@ -105,6 +105,7 @@ def _index_units(year, month, day, unit="day", respect_leap_years=True) -> (np.a
 def calc_climatology(Ser,
                      moving_avg_orig=5,
                      moving_avg_clim=35,
+                     moving_avg_month_clim=3,
                      median=False,
                      timespan=None,
                      fill=np.nan,
@@ -128,9 +129,14 @@ def calc_climatology(Ser,
         Default: 5
 
     moving_avg_clim : float, optional
-        The size of the moving_average window [units] that will be applied on the
+        The size of the moving_average window in days that will be applied on the
         calculated climatology (long-term event correction)
         Default: 35
+
+    moving_avg_month_clim: : float, optional
+        Same as for 'moving_avg_clim', but applied to monthly climatologies. In case
+        unit='month', this value overrides 'moving_avg_clim'
+        Default: 3
 
     median : boolean, optional
         if set to True, the climatology will be based on the median conditions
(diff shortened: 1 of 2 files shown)


src/pytesmo/time_series/anomaly.py
```diff
@@ -87,9 +87,24 @@ def calc_anomaly(Ser,
...
+        args = month, day, year
+    else:
+        args = month, day
+
+    idx_lut = {
+        "day": (doy(*args), 366),
+        "month": (month, 12),
+    }
+
+    return idx_lut[unit]
```
s-scherrer, line 102: Can you reformulate this as an if-else statement?
```
if unit == "day":
    return doy(*args), 366
elif unit == "month":
    return month, 12
else:
    raise ValueError(f"Invalid unit: {unit}")
```

This way the function `doy` is only called if it is really needed.

src/pytesmo/time_series/anomaly.py
```diff
@@ -205,8 +229,9 @@ def calc_climatology(Ser,
     else:
         clim_ser = moving_average(clim_ser, window_size=moving_avg_clim, fillna=fillna, min_obs=min_obs_clim)
 
-    clim_ser = clim_ser.reindex(np.arange(366) + 1)
+    clim_ser = clim_ser.reindex(np.arange(n_idx) + 1)
```
s-scherrer, line 232: Can you move this line before the `if wraparound` block? This would solve one of the issues in #250

src/pytesmo/time_series/anomaly.py
```diff
@@ -87,9 +87,24 @@ def calc_anomaly(Ser,
...
+    return idx_lut[unit]
+
+
 def calc_climatology(Ser,
                      moving_avg_orig=5,
-                     moving_avg_clim=30,
+                     moving_avg_clim=35,
                      median=False,
                      timespan=None,
                      fill=np.nan,
```
s-scherrer, line 110: Can you change the line below to `wraparound=True`? This solves #250

s-scherrer (approved):

src/pytesmo/time_series/anomaly.py
```diff
@@ -87,19 +87,35 @@ def calc_anomaly(Ser,
...
+    elif unit == "month":
+        return month, 12
+    else:
+        raise ValueError(f"Invalid unit: {unit}")
+
+
 def calc_climatology(Ser,
                      moving_avg_orig=5,
-                     moving_avg_clim=30,
+                     moving_avg_clim=35,
```
s-scherrer, line 107: For unit="month" 35 is not a good default value, this should probably be checked at the start of the function, something like
```
if unit == "month" and moving_avg_clim == 35:
    moving_avg_clim = <default value>
```
where the default value could be 1, 2, or 3.

s-scherrer: This should then also solve the failing tests

pstradio: `test_climatology_always_366_fill` and `test_climatology_interpolate_leapday` fail because we are now reindexing `clim_ser` in line 223, i.e. before the `moving_average()` and this leads somehow to a non-gapped climatology. I think these tests are obsolete as they are, but not sure how they should be updated

s-scherrer: For `test_climatology_always_366_fill` you can set `moving_avg_clim=1`. This way there will still be gaps in the climatology, I think. I would remove the other test actually, because the behaviour that it asserts does not make a lot of sense in my opinion.

Issue #251, opened 2021-11-26
Skip Triple Collocation CIs when not enough data is available instead of raising Error


wpreimes:
Metrics Operator for TC metrics with CI raises an error when less than 100 observations (default) are available https://github.com/TUW-GEO/pytesmo/blob/master/src/pytesmo/validation_framework/metric_calculators.py#L1740

I think it would make more sense if in those cases the CIs are skipped (e.g. NaN) but the TC results are still returned. @s-scherrer @pstradio @sheenaze?

sheenaze:
Does it make sense to return only TC? Wouldn't be an inconsistency? I mean... you're the one who knows it better, the most important thing is to handle it, so it doesn't crash the entire celery job.

s-scherrer:
Closed via #256

PR #252, opened 2021-11-29
Only raise warning when not enough data for bootstrapping

s-scherrer:
I changed the bootstrapping functions to warn and return NaN for the confidence intervals instead of raising an error when there is not enough data to do bootstrapping, as discussed in #251

### DIFF
--- a/src/pytesmo/metrics/confidence_intervals.py
+++ b/src/pytesmo/metrics/confidence_intervals.py
@@ -1,5 +1,6 @@
 import numpy as np
 from scipy import stats
+import warnings
 
 from pytesmo.metrics import pairwise
 from pytesmo.metrics.tcol import tcol_metrics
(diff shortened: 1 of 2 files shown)


s-scherrer:
This issue again:

```
  File "/usr/share/miniconda3/envs/pytesmo/bin/pytest", line 8 in <module>
tests/test_validation_framework/test_validation.py::test_ascat_ismn_validation_metadata_rolling 
/home/runner/work/_temp/2c06d129-8349-4c2d-a0df-7358a0b65a24.sh: line 4:  4925 Segmentation fault      (core dumped) pytest --cache-clear -m 'full_framework'
```

I think it's unrelated to my changes, since the tests run on 3.7, but I thought that running the full_framework tests separately solved the issue.

wpreimes:
> This issue again:
> 
> ```
>   File "/usr/share/miniconda3/envs/pytesmo/bin/pytest", line 8 in <module>
> tests/test_validation_framework/test_validation.py::test_ascat_ismn_validation_metadata_rolling 
> /home/runner/work/_temp/2c06d129-8349-4c2d-a0df-7358a0b65a24.sh: line 4:  4925 Segmentation fault      (core dumped) pytest --cache-clear -m 'full_framework'
> ```
> 
> I think it's unrelated to my changes, since the tests run on 3.7, but I thought that running the full_framework tests separately solved the issue.

I think it did ... until a few weeks ago. I will take a look at this today and try to fix it.

wpreimes:
btw, the reason why I have not merge this yet is that I should implement an option into the validation framework that allows a user to skip points where an error during metric calculation occurred.
E.g. An option to continue on errors in this loop: https://github.com/TUW-GEO/pytesmo/blob/master/src/pytesmo/validation_framework/validation.py#L239
In that case it would be fine if the Bootstrapping metrics are only calculated when the bootstrapping is possible (which would be more consistent, otherwise we have more points where the metric was calculated than points where the CI was calculated), right?

s-scherrer:
Right, that's probably the better solution, I didn't think of this before. Let's add the error handling in this loop then.

PR #256, opened 2021-12-16
Calc debugging

pstradio:
Handles exceptions in `validation_framework.validation.calc()` when a metric is not computed (e.g. TC errors) and not written to the results dictionary.

### DIFF
--- a/src/pytesmo/validation_framework/validation.py
+++ b/src/pytesmo/validation_framework/validation.py
@@ -265,7 +265,14 @@ class Validation(object):
             for field_name in results[key][0].keys():
                 entries = []
                 for result in results[key]:
-                    entries.append(result[field_name][0])
+                    try:
+                        entries.append(result[field_name][0])
+                    except KeyError:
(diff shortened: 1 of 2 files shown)


src/pytesmo/validation_framework/validation.py
```diff
@@ -265,7 +265,15 @@ def calc(
             for field_name in results[key][0].keys():
                 entries = []
                 for result in results[key]:
-                    entries.append(result[field_name][0])
+                    try:
+                        entries.append(result[field_name][0])
+                    except KeyError as e:
+                        # indicating a metric result was not produced
+                        warnings.warn(
+                            f"No TCA results for gpi: {result['gpi']}. Error "
```
s-scherrer, line 273: I would write something like f"No results for {key} for gpi: {...}" so the error message is also correct if the issue is not TCA related. Might lead to less confusion in case something else but TCA fails, especially during development.

s-scherrer (approved):

PR #260, opened 2022-01-12
Option to ignore errors in validation framework, use new ISMN interface in tests

wpreimes:
Option to ignore errors in validation framework, use new ismn interface in tests, fix binary incompatibilities using oldest-supported-numpy, drop py 3.6 (as this is not supported by latest pytest and other dependencies).


wpreimes:
also see:
https://github.com/scipy/oldest-supported-numpy
https://stackoverflow.com/a/66743692/17376826
https://github.com/scikit-learn-contrib/hdbscan/issues/457

wpreimes:
I should split this up into at least 2 PRs... will do that. @s-scherrer but to fix the `pip install` I think really only https://github.com/wpreimes/pytesmo/blob/master/pyproject.toml#L2 is relevant.

sebhahn:
Does it mean the next pytesmo version will be >= py 3.7?

s-scherrer:
> I should split this up into at least 2 PRs... will do that. @s-scherrer but to fix the `pip install` I think really only https://github.com/wpreimes/pytesmo/blob/master/pyproject.toml#L2 is relevant.

Super, thanks a lot for fixing this!

wpreimes:
> Does it mean the next pytesmo version will be >= py 3.7?

from 0.14 on yes, to fix the pip installation I would release 0.13.4 with python 3.6 as well

sebhahn:
would pip install work if pytest is pinned?

s-scherrer (approved):

wpreimes:
as long as pip does not run into version conflicts (with packages defined in setup.cfg) it should still work with old python versions as well.

PR #262, opened 2022-01-13
fixed bugs in resample_mean

s-scherrer:
This should fix the bugs that were found in the documentation testcase.

The problems were due to indexing errors at the start and the end of the arrays. I will add a few tests tomorrow.


wpreimes:
Thanks, looks good to me.

PR #266, opened 2022-04-04
Metrics for temporal subgroups

wpreimes:
Add a more flexible wrapper to calculate metrics for temporal subsets, e.g to compute metrics using data for certain days of a year. We used this one e.g. in SMART-DRI to calculate correlation values for groups of 10-days (over 13 years) to estimate varying levels of subsurface scattering throughout the year (wet/dry season).

Before, this was only possible on a monthly basis: https://github.com/TUW-GEO/pytesmo/blob/master/src/pytesmo/validation_framework/metric_calculators_adapters.py#L43

Now there is a [data distributor ](https://github.com/wpreimes/pytesmo/blob/set-metrics/src/pytesmo/validation_framework/metric_calculators_adapters.py#L123) to extract certain dates, date_ranges for any year to apply the metrics calculator to.
However, this might be over-complicated, but I haven't found a way to only use e.g. pandas for this yet, but I'm pretty sure it is possible using groupby or so.

@pstradio @s-scherrer if you have any comments let me know. If we find no better solution I will probably add [this class ](https://github.com/wpreimes/pytesmo/blob/set-metrics/src/pytesmo/validation_framework/metric_calculators_adapters.py#L53) to [cadati](https://github.com/TUW-GEO/cadati) @sebhahn


sebhahn:
@wpreimes yes, I'd support to move this class to cadati

Is there not way to inherit from a standard python datetime object or similar?

tests/test_validation_framework/test_metric_calculators_adapters.py
```diff
@@ -0,0 +1,142 @@
...
+    def test_doy(self):
+        assert GenericDatetime.from_datetime(self.future).doy == 159
+        try:
+            _ = datetime(self.now.year, 2, 29)
+            assert self.now.timetuple().tm_yday == self.gnow.doy
+        except ValueError:
+            if self.now > datetime(self.now.year, 2, 28):
+                assert self.now.timetuple().tm_yday + 1 == self.gnow.doy
+            else:
+                assert self.now.timetuple().tm_yday == self.gnow.doy
```
s-scherrer, line 37: In general I think it's better if we set a few dates to test instead of using `datetime.now`, so that the actual test code that is run is the same all the time.

Could you also add a comment about why the try except statement is necessary, and for the if-else statement?

src/pytesmo/validation_framework/metric_calculators_adapters.py
```diff
@@ -24,73 +24,297 @@
...
+            return dt
+
+
+class TsDistributor:
+
+    def __init__(self,
+                 dates=None,
+                 date_ranges=None,
+                 generic_dates=None,
+                 generic_date_ranges=None):
```
s-scherrer, line 129: I'm not a fan of long keyword argument names, they are always hard to remember. I would instead use
1) 2 keywords, `dates` and `date_ranges`, and do type/attribute checks to figure out whether it's a datetime or a GenericDatetime
2) or accept only a single list of individual datetimes, GenericDatetimes, or tuples of datimes and GenericDatetimes.

As a more general comment about design:

From the tests and how you used the class in the code I also got the impresssion that only the `select` method of the class is really supposed to be used. Since the class doesn't really have a state it might be conceptually simpler to make a function out of this (e.g. select_subset).

If we implement option 2 from above, we could also simplify the interface of SubsetsMetricAdapter to accept a dictionary of lists instead of a dictionary of TsDistributors, and make SubsetsMetricAdapter generate the TsDistributors/call select_subset using the lists as arguments. This would make SubsetsMetricAdapter a bit easier to use.

tests/test_validation_framework/test_metric_calculators_adapters.py
```diff
@@ -0,0 +1,142 @@
...
+        self.now = datetime.now()
+        self.past = datetime(1900, 1, 2, 3, 4, 5)
+        self.future = datetime(2104, 6, 7, 8, 9, 10)
+        self.gnow = GenericDatetime(self.now.month, self.now.day,
+                                    self.now.hour, self.now.minute,
+                                    self.now.second)
+
+    def test_comparisons(self):
+        assert self.gnow > GenericDatetime.from_datetime(self.past)
+        assert self.gnow < GenericDatetime.from_datetime(self.future)
```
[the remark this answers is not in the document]
wpreimes: yeah, that doesnt make much sense I think ;) I will change that

tests/test_validation_framework/test_metric_calculators_adapters.py
wpreimes: agree, will update it

wpreimes:
@s-scherrer thanks for the feedback. I will update this PR. I will also add an small example notebook, which might clear up a few things.

src/pytesmo/validation_framework/metric_calculators_adapters.py
s-scherrer: Okay, then we just have different preferences on this :D You can also leave it then, just an idea I had when going through the code.

wpreimes:
I have implemented the above suggested changes. There is now an example in the the notebook at `pytesmo/docs/examples/validation_framework.ipynb`. The `GenericDatetime` has been renamed to `YearlessDatetime` which better reflects the purpose. However, I will keep this in pytesmo for now (`pytesmo/time_series/grouping.py`) as this is very specific and I currently don't see any usage outside of pytesmo.

PR #267, opened 2022-04-14
Make bootstrapping settings better accessible when using the validation framework

wpreimes:


wpreimes:
We are experimenting with smaller bootstrapping samples and different CI levels in our validation framework. To make it possible to these settings from outside of pytesmo I have added the options to the Metric calculator classes.

src/pytesmo/metrics/confidence_intervals.py
```diff
@@ -328,15 +328,15 @@ def _jackknife(metric_func, x, y):
 
 def _percentile_bs_ci(bs_m, m, alpha):
     """Calculates the CI using percentiles"""
-    lower = np.quantile(bs_m, alpha / 2)
-    upper = np.quantile(bs_m, 1 - alpha / 2)
+    lower = np.nanquantile(bs_m, alpha / 2)
```
wpreimes, line 331: is it expected that there can be nans when using small samples?

s-scherrer: I think that the metric functions should normally not return NaN, but it's probably safer to use nanquantile

s-scherrer (approved):

PR #282, opened 2023-01-27
Ignore nans masking

pstradio:
Method for masking to ignore NaNs in filter columns, i.e. observations can be kept if no information is given by the filtering field.

### DIFF
--- a/src/pytesmo/validation_framework/adapters.py
+++ b/src/pytesmo/validation_framework/adapters.py
@@ -91,8 +91,8 @@ class BasicAdapter:
 
     def __get_dataframe(self, data):
         if ((not isinstance(data, DataFrame)) and
-            (hasattr(data, self.data_property_name)) and
-            (isinstance(getattr(data, self.data_property_name), DataFrame))):
+                (hasattr(data, self.data_property_name)) and
+                (isinstance(getattr(data, self.data_property_name), DataFrame))):
             data = getattr(data, self.data_property_name)
         return data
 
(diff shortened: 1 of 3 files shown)


wpreimes:
I think @sebhahn removed some functions from pygeobase recently, which are not found anymore by pytesmo...
https://github.com/TUW-GEO/pygeobase/compare/v0.4.0...v0.5.0#diff-59daaa55512c1966afb3f137b3c4d21b53f024252068ddd7b0ee377f27669d4eL683
for a quick fix you could try to fall back to pygeobase==0.4.0

pstradio:
That woul

> I think @sebhahn removed some functions from pygeobase recently, which are not found anymore by pytesmo... [TUW-GEO/pygeobase@v0.4.0...v0.5.0#diff-59daaa55512c1966afb3f137b3c4d21b53f024252068ddd7b0ee377f27669d4eL683](https://github.com/TUW-GEO/pygeobase/compare/v0.4.0...v0.5.0#diff-59daaa55512c1966afb3f137b3c4d21b53f024252068ddd7b0ee377f27669d4eL683) for a quick fix you could try to fall back to pygeobase==0.4.0
> 
That would work..I will just fallback for now and open an issue

pstradio:
@s-scherrer could you please review and merge if you think it is ok? new tests are included but for some reason coverall is not happy

src/pytesmo/validation_framework/adapters.py
```diff
@@ -309,14 +314,18 @@ def _adapt(self, data):
...
+            else:
+                new_mask = operator(data[column_name], threshold)
 
             if mask is not None:
                 mask = mask & new_mask
             else:
                 mask = new_mask
 
-        return data[mask]
+        return data[mask].dropna(how="all")
```
s-scherrer, line 328: What's the reason for adding the `dropna` here?

pstradio: With the `ignore_nans` option active, the dataframe can be returned with all-nans rows. I think it is quite specific to the SMOSL2 data, where there are a LOT of NaNs in the flagging fields, and a lot of observation with no soil moisture (NaN) and some type of flag value.

In the latter case it can happen that the NaN soil moisture is kept. `dropna` just eliminates these occurrences

s-scherrer: Sounds okay to me. I will merge it in then. Do you need a new release?

pstradio: I guess it would be best, but if you prefer to wait it should work too.

PR #288, opened 2023-02-20
Fix tests, CI, environment.yml

wpreimes:


wpreimes:
@s-scherrer I have now simply excluded netcdf4 v1.6.2. But if they don't fix the issue in the next version, the error will come back again. Switching the channels in the yml file (to use defaults over conda-forge) would also solve this issue, without excluding the netcdf4 version even.  `pyresample` is **only** available on conda-forge, so we have to include the conda-forge channel (at least as a fallback option).

PR #290, opened 2023-02-27
Add std kwarg to anomaly function

wpreimes:


wpreimes:
Returning the std dev from CLIM computation would allow using it to e.g. standardize anomalies

wpreimes:
I still need to check whether smoothing std.dev makes sense for daily data, and add tests

